PR #17, opened 2022-03-28
User authentication

aronhoyer:


ovineq (changes requested):
Went through the code with Aron.

Decided on removing some of the unused environment variables and use the http client he added instead of `fetch` calls.

ovineq (approved):
LGTM 🎉

PR #20, opened 2022-04-26
Frontpage UI

aronhoyer:


ovineq (approved):
Nice work 🥳 
The front page is starting to look good and the code is easy to follow!

Suggestion: Could we add an entry to the npm scripts that generates the mock data? Maybe a comment on the README aswell? 😇 

I see we'll need to update `package-lock.json` with the optional `@next/swc-{version}`s we'll want to support in the future (devtops/CI). I'll add what I need for my linux setup once I start contributing code.

Issue #23, opened 2022-05-04
Add github actions integration checks


ovineq:
Building off from previous projects integration checks.
Nextjs runs eslint as part of its build process, so we won't need a seperate step for that. Test will come in ones we stabilize a little more.

PR #24, opened 2022-05-04
Add build workflow

ovineq:
Linting is part of Nextjs build process, so we'll skip that for now.
A step for tests can also come later.

Issue: https://github.com/equinor/data-marketplace/issues/23

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,0 +1,58 @@
+name: Checks
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yaml
```diff
@@ -0,0 +1,58 @@
+name: Checks
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
```
aronhoyer, line 6: should be `main`

aronhoyer (changes requested):

.github/workflows/checks.yaml
ovineq: Copy-paste error. Fixed ✔️

PR #34, opened 2022-05-24
Search page

### LINKED ISSUE #33: Search view and backend

aronhoyer:
This pull request adds search.

As of now, search is tightly coupled with Collibra.

New:
- API endpoint to forward search keyword and user selected filters to Collibra search
- Search page UI
  - Search results
  - Filters
    - Filters are added to the URL as search parameters, so a search results are sharable

Resolves #33


components/AssetCard/AssetCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <Typography variant="h4" as="p">{title}</Typography>
+        </div>
+
+        {meta && (
+          <MetaContainer>
+            {meta.map((m, i) => (
+              <Typography variant="caption" key={`${id}-meta-${i + 1}`}>
+                {m.label}
+                {" "}
+                {m.value}
```
ovineq, line 63: A suggestion:
```
[m.label, m.value].join(" ")
```

pages/search.tsx
```diff
@@ -0,0 +1,238 @@
...
+    } else {
+      const existingIdx = community.findIndex((f) => f === id)
+      if (existingIdx === -1) {
+        community.push(id)
+      } else {
+        community = [
+          ...community.slice(0, existingIdx),
+          ...community.slice(existingIdx + 1, community.length),
+        ]
+      }
```
ovineq, line 147: An alternative approach since we only use the index for existence check:
```
      const filteredCommunity = community.filter((f) => f !== id)
      if (filteredCommunity.length === community.length) {
         community.push(id)
      } else {
        community = filteredCommunity
      }
```

ovineq (changes requested):
I see we're adding more logic now. Perhaps it's time to include some simple tests for the logic flows we have?

components/AssetCard/AssetCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </MetaContainer>
+        )}
+
+        <div>
+
+          {/*
+            * html strings are sanitized in backend at this point,
+            * so we should(TM) be safe to do this
+            */}
+          <TruncatedDescription variant="body_long" dangerouslySetInnerHTML={{ __html: description }} />
```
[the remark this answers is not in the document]
aronhoyer: it's sanitized on our backend

https://github.com/equinor/data-marketplace/pull/34/files#diff-6e45d42774ee51840cac5ce12248a073517576922f45c3adfbc4822cee2bca6bR57

components/AssetCard/AssetCard.tsx
aronhoyer: that introduces unnecessary array operations, though

pages/search.tsx
aronhoyer: we also use the index to be able to remove an item at the given index. if a user has clicked a community in the filter list, they should be able to remove it by clicking it again.

pages/search.tsx
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    router.push({
+      pathname: router.pathname,
+      query: { ...router.query, community },
+    })
+  }
```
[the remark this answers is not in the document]
aronhoyer: updated and tested! 😄

pages/api/search.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      res.json({
+        ...searchRes.body,
+        results,
+      })
+    } catch (error) {
+      // console.log("[SearchHandler]", error)
+
+      const err = error as HttpError
+      res.status(err.statusCode ?? 500).send(err.body)
```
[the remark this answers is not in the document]
ovineq: Sounds good 👍🏻

pages/search.tsx
ovineq: Yup, the suggested codeblock also covers that without needing the index.
Nice work with the typedoc and tests 👏🏻

PR #36, opened 2022-05-24
Expose public path in workflow env

ovineq:
In order for next to build on github actions

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -42,8 +42,8 @@ jobs:
           key: ${{ github.sha }}
   build:
     name: Build
-    runs-on: ubuntu-latest
     needs: setup
+    runs-on: ubuntu-latest
     steps:
       - name: Get base files
         uses: actions/cache@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/checks.yaml
```diff
@@ -56,3 +56,5 @@ jobs:
         run: |
           npm run seed:mock:all
           npm run build
+        env:
```
aronhoyer, line 59: doesn't this need to be set **before** `run`? 🤔

ovineq: Didn't think it mattered as long as it's in the same scope/indentation. But testing now. 👍🏻

ovineq:
> I feel like we should find a different solution to this and `.env` should be added to `.gitignore`. Allowing the file to be pushed to the remote opens up for someone accidentally committing their env config.
> 
> @nilsml had some experience with reading next js env vars from a different project, so maybe link up with him smile

Removed the `.env` and added it to the ignore list. 
Using localhost as the environment variable works, and should suffice for this workflow.
We can revisit the usage of secrets when we're setting up environments.

aronhoyer (approved):
looks great!

PR #39, opened 2022-06-16
feat: Add asset view

### LINKED ISSUE #30: Implement asset view

aronhoyer:
This pull request adds UI for the asset view and some additional types for the Collibra API.

Resolves #30


pages/api/assets/[id]/responsibilities.ts
```diff
@@ -25,20 +25,37 @@ const AssetResponsibilitiesHandler: NextApiHandler = async (req, res) => {
...
 
-          return {
+          if (!collibraUser) return obj
+
+          const transformedUser = {
             ...user,
-            firstName: fullUser.firstName,
-            lastName: fullUser.lastName,
-            email: fullUser.emailAddress,
+            firstName: collibraUser.firstName,
```
Comment (nilsml) [inline pages/api/assets/[id]/responsibilities.ts:39]: A small improvement to this code is to destruct cloibraUser first and then use the result in transformedUser:

```
const { firstName, lastName, emailAddress: email } = collibraUser
const transformedUser = { ...user, firstName, lastName, email }
```

collibra.d.ts
```diff
@@ -149,5 +188,6 @@ declare namespace Collibra {
   export interface PagedAttributeResponse extends PagedResponse<Attribute> {}
```
nilsml, line 188: Open question for the team: should we have both interfaces and types in our codebase? In many projects I have seen only type to be used in order to have less confusion.

collibra.d.ts
```diff
@@ -119,19 +115,62 @@ declare namespace Collibra {
...
+    city: string
+    street: string
+    number: string
+    state: string
+    country: string
+    postalCode: string
+    type: "HOME" | "WORK"
+  }
+
+  export type User = Resource & {
```
nilsml, line 153: Good use of types!

nilsml (commented):
Code looks nice and clean. Probably someone else should give thumb up on the CSS part.
One general note is that this is a pretty big PR. But as long as it is for one feature I guess it is ok. Smaller PR's are easier to review ;)

collibra.d.ts
aronhoyer: I read an article once about what the difference is and when which should be used. essentially, the author of the article said that the way they read types and interfaces, was that interfaces are like classes—made to be extended. when they saw an interface, they expected it to get extended. meanwhile types were more static.

however, functionally they're the same. i also believe types don't accept generics.

nilsml: Yes, that is true. I just know that some developers get a bit confused on what to use where, so by taking one away the choice ise easier. To me it is more about distancing myself from object oriented programming :smile

pages/api/assets/[id]/responsibilities.ts
Comment (aronhoyer) [reply pages/api/assets/[id]/responsibilities.ts:39]: while I agree to a certain extent, in my opinion, destructuring a lot of properties makes the code messier\

collibra.d.ts
```diff
@@ -119,19 +115,62 @@ declare namespace Collibra {
...
     baseResource: ResourceReference
     owner: ResourceReference
   }
 
+  type Email = Resource & {
+    emailAddress: string
+  }
+
+  type PhoneNumber = Resource & {
+    type: "FAX" | "MOBILE" | "OTHER" | "PAGER" | "PRIVATE" | "WORK"
```
wenche, line 129: Fax 😅

pages/assets/[id].tsx
```diff
@@ -1,71 +1,167 @@
...
+                Add to cart
+              </Button>
+            </>
+          )}
+        </Header>
+
+        <Divider />
+
+        <ContentContainer>
+          <TabNavContainer>
```
Comment (wenche) [inline pages/assets/[id].tsx:147]: I guess I have to ask... Did you consider to use the eds tabs? That would also give us the correct accessibility information that is needed for tabs.

Comment (aronhoyer) [reply pages/assets/[id].tsx:147]: I guess we could do that. I only worry that the ui becomes very "vertical" if that makes sense. I could make a quick Figma sketch that uses the eds tabs

components/AssetTabContent/AssetTabContentSectionContainer.tsx
```diff
@@ -0,0 +1,11 @@
+import styled from "styled-components"
+
+export const AssetTabContentSectionContainer = styled.section`
+  &:not(:last-child) {
+    margin-bottom: 1.5rem;
+  }
+
+  h2 {
+    margin-bottom: 0.5rem;
```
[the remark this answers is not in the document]
aronhoyer: why is that?

pages/assets/[id].tsx
Comment (wenche) [reply pages/assets/[id].tsx:147]: It's quite a huge change I guess. But if we don't do that, we would at least have to deal with some a11y things ourselves. We could put a card to the backlog and investigate this further, as there is probably more components that should ideally utilise EDS? As a side note, we didn't use the EDS tabs in Envis as well due to a lot of style changes instead we used reach tabs. :)

components/AssetTabContent/AssetTabContentSectionContainer.tsx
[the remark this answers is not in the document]
aronhoyer: the way I see it a given component shouldn't set its spacings, since that often times depends on the context in which it's used. but if we were to set the rules for spacing for each component, and especially for typography, the rule set can become extremely large.

pages/assets/[id].tsx
Comment (aronhoyer) [reply pages/assets/[id].tsx:147]: I've drawn up a quick [Figma sketch](https://www.figma.com/file/54wzCBdXkEbWsAGXOVWLd6/?node-id=128%3A1318) that I can link to in the ticket.

The design works in my opinion, but as you said, it's not a trivial change.

wenche (approved):
LGMT :thumbsup: Nice work

PR #49, opened 2022-06-18
refactor: Add backend for fetching communities

### LINKED ISSUE #47: Get Equinor community ID from Collibra

aronhoyer:
Resolves #47

### DIFF
--- a/pages/api/communities.ts
+++ b/pages/api/communities.ts
@@ -1,0 +1,35 @@
+import { NextApiHandler } from "next"
+
+import { config } from "../../config"
+import { HttpClient } from "../../lib/HttpClient"
+
+const CommunitiesHandler: NextApiHandler = async (req, res) => {
+  switch (req.method) {
+    case "GET": {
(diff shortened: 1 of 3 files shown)


pages/api/communities.ts
```diff
@@ -0,0 +1,35 @@
...
+        query: {
+          name: "equinor",
+          nameMatchMode: "ANYWHERE",
+        },
+      })
+
+      const equinorCommunity = allComunities.body?.results[0]
+
+      if (!equinorCommunity) {
+        return res.status(500).end()
```
nilsml, line 20: Why would this occur? Could it be that there are no communities? In case 404 would be correct?

nilsml (commented):
Looks good, but one question about return code.

pages/api/communities.ts
aronhoyer: if the Equinor community doesn't exist, it would probably mean that something changed in Collibra. in that case a 404 would not be the correct return code, as what the consumer is asking for is not the Equinor community, but the sub-communities of the Equinor community.

PR #51, opened 2022-06-21
Utilize i18n package for static texts

### LINKED ISSUE #16: Utilize i18n package for static text
To make static texts easier to maintain, and potential translations down the line easier to implement, we should add a package like [next-i18next](https://www.npmjs.com/package/next-i18next).

This will allow us to store static text in one place, which will make it easier to maintain copy.

### Acceptance criteria
- All texts are added to a file for English text
- The components use these texts instead of hard coded text inside the code itself

wenche:
Resolves #16


locales/en.json
```diff
@@ -0,0 +1,4 @@
+{
+  "frontpage_documentTitle": "Data Marketplace",
+  "frontpage_popularProductsHeader": "Popular"
+}
```
aronhoyer, line 4: I would like for us to be able to namespace translations. I see this getting cumbersome to maintain at scale >.<

wenche: By changing library or by using the IntlProvider in the pages instead? Might perhaps be easier with another library 🤔 If we at some point want to remove the string to outside of the code base, we can solve that problem when it arises

aronhoyer: unless into-react has a limitation, couldn't we namespace by file?

aronhoyer (approved):
lgtm 👍

PR #62, opened 2022-06-24
feature/WT-60-update-EDS-packages

### LINKED ISSUE #60: Upgrade EDS packages

wenche:
Resolves #60


wenche:
The focus-visible polyfill was used by EDS before the native browser support improved. It is no longer needed and EDS has already updated their code to use the native :focus-visible selector instead of the data attribute that  this polyfill package adds.

aronhoyer (approved):

PR #79, opened 2022-07-25
feat: Add checkout access view

### LINKED ISSUE #80: Create "Request access" step for checkout wizard
[Figma sketch](https://www.figma.com/file/54wzCBdXkEbWsAGXOVWLd6/?node-id=2%3A11826)

User should not be able to go to the next step before filling out a description.

The description should not exceed a character limit (TBD).

aronhoyer:
Resolves #80

### DIFF
--- a/pages/checkout/access.tsx
+++ b/pages/checkout/access.tsx
@@ -1,0 +1,107 @@
+import { Button, TextField, Typography } from "@equinor/eds-core-react"
+import type { NextPage } from "next"
+import { useRouter } from "next/router"
+import { ChangeEventHandler } from "react"
+import { useIntl } from "react-intl"
+import { useDispatch, useSelector } from "react-redux"
+import styled from "styled-components"
+
+import { CheckoutWizard } from "../../components/CheckoutWizard/CheckoutWizard"
+import { Container } from "../../components/Container"
+import { Dispatch, RootState } from "../../store"
+
+const Headline = styled(Typography).attrs(() => ({ variant: "h4" }))`
+  margin-bottom: 0.25rem;
+`
+
+const Ingress = styled(Typography).attrs(() => ({ variant: "ingress" }))`
+  margin-bottom: 2rem;
+`
+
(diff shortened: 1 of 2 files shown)


pages/checkout/access.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { ChangeEventHandler } from "react"
+import { useIntl } from "react-intl"
+import { useDispatch, useSelector } from "react-redux"
+import styled from "styled-components"
+
+import { CheckoutWizard } from "../../components/CheckoutWizard/CheckoutWizard"
+import { Container } from "../../components/Container"
+import { Dispatch, RootState } from "../../store"
+
+const Headline = styled(Typography).attrs(() => ({ variant: "h4" }))`
```
wenche, line 13: Personally I find it easier to read if the variant is used as a prop on the component itself 🤷‍♀️

pages/checkout/access.tsx
```diff
@@ -0,0 +1,107 @@
...
+`
+
+const TextFieldContainer = styled.div`
+  margin-bottom: 1.5rem;
+`
+
+const CheckoutAccessView: NextPage = () => {
+  const intl = useIntl()
+  const dispatch = useDispatch<Dispatch>()
+  const description = useSelector((state: RootState) => state.checkout.data.access?.description) ?? ""
```
wenche, line 46: A good use case for the nullish operator :)

wenche (approved):
LGTM 👍 

We should at some point look at the pattern for disabling the submit button vs proper validation. I very much prefer the pattern of always having the submit button enabled, and rather let the form validation restrict you from submitting. Easier to understand for everyone IMHO Let's discuss this with Trond as well :)

aronhoyer:
> I very much prefer the pattern of always having the submit button enabled, and rather let the form validation restrict you from submitting.

Very true! I've actually had that discussion with a different team, so I'm honestly not sure why I designed and implemented it this way 😅

PR #90, opened 2022-07-27
Feature/nlun 56 set up hosting

nilsml:
First version of dockerfile and radixconfig

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM node:lts-alpine AS dependencies
+WORKDIR /opt/app
+COPY package.json package-lock.json ./
+RUN npm ci
+
+
+FROM node:lts-alpine AS builder
+WORKDIR /opt/app
+COPY . .
+COPY --from=dependencies /opt/app/node_modules ./node_modules
+RUN npm run build
+
+
+FROM node:lts-alpine AS runner
+WORKDIR /opt/app
+ENV NODE_ENV production
+ENV PORT 3000
+ENV USER nextjs
+ENV UID 12345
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM node:lts-alpine AS dependencies
+WORKDIR /opt/app
+COPY package.json package-lock.json ./
+RUN npm ci
```
wenche, line 4: npm ci ❤️

Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM node:lts-alpine AS dependencies
```
wenche, line 1: I guess lts is what we must do to be kind to future self, but let's just be aware of this 🕵️‍♀️

radixconfig.yaml
```diff
@@ -0,0 +1,29 @@
...
+    environmentConfig:
+    - environment: prod
+      resources:
+        requests:
+          memory: 512Mi
+          cpu: 100m
+        limits:
+          memory: 1Gi
+          cpu: 1000m
+      variables:
```
wenche, line 23: I guess this is correct, but it just looks funny with two variables following each other

wenche (approved):
LGTM 👍

radixconfig.yaml
nilsml: Yeah, could omit them, but we will soon be using them

Dockerfile
nilsml: Yes. Maybe a little overkill right now, but still the file is quite easy to read and understand I think

PR #98, opened 2022-07-29
👷 Fix: 97 Log errors to Slack from Github Action

nilsml:
If something fails while running github action, this will be reported to the ci channel


wenche (approved):
LGTM 👍

PR #96, opened 2022-07-29
Use real data for the cart content

wenche:
Resolves #93

PR #104, opened 2022-08-01
Cart view with real data

wenche:
resolves #96


nilsml (approved):
Discussed in person a few enhancements that can be done in the future, but all in all this looks good.

PR #105, opened 2022-08-01
Feature/sm 71 redirectto frontpage

ShikhaMishra9:


components/NavBar/NavBar.tsx
```diff
@@ -125,18 +125,22 @@ export const NavBar: VoidFunctionComponent = () => {
       <Header>
         <HeaderContentContainer>
           <LogoContainer>
-            <Typography><FormattedMessage id="navbar.logo" /></Typography>
+            <Link href="/">
+              <Button variant="ghost" color="secondary">
```
wenche, line 129: Since this is a button used for navigation, I think they should be a real link `a` and not a `button`. I like to think of them as a link styled like a button. :)

With the EDS button, you can add a prop `as="a"` and then it will render as a link, not a button.

wenche (changes requested):
Looks good, but let's stick to links for navigation and buttons for actions :)

wenche (approved):
LGTM 👍

PR #112, opened 2022-08-03
Adding banner component

ShikhaMishra9:


components/Banner/Banner.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const Banner: FunctionComponent<Props> = ({ variant, children }) => {
+  const classNames = [
+    "banner",
+    variant === "danger" ? "banner--danger" : "banner--warning",
+  ]
+    .join(" ")
+
+  return (
+    <div className={classNames}>
```
wenche, line 19: Pay attention to the centering and alignment of the icon and text in the Figma sketches. The icon should not be centered along the y-axis if we have several paragraphs of text, but if we have only one line of text both the icon and text should be centered. Now it is a little bit off.
Skjermbilde 2022-08-03 kl  10 01 16

pages/cart.tsx
```diff
@@ -111,7 +110,9 @@ const CartView : NextPage = () => {
           </CartItems>
         )}
 
-        <p style={{ padding: "1rem", margin: "1rem 0", backgroundColor: `${tokens.colors.ui.background__warning.rgba}` }}>Placeholder for a banner component</p>
+        <Banner variant="warning">
+          {intl.formatMessage({ id: "cart.banner.warning" })}
```
wenche, line 114: We need to wrap this in the Typography component to get the font correct and also to get a `<p>` tag

pages/checkout/terms.tsx
```diff
@@ -65,7 +77,14 @@ const CheckoutTermsView: NextPage = () => {
         </IngressContainer>
 
         {/* TODO: Add banner */}
-
+        <BannerContainer>
+          <Banner variant="danger">
+            <Title>{intl.formatMessage({ id: "terms.banner.danger.heading1" })}</Title>
```
wenche, line 82: Add some semantics to the Title here using the Typography component. It's always important to use the Typography component as it adds Equinor styles to the text

styles/banner.css
```diff
@@ -0,0 +1,40 @@
+.Cart {
+  font-family: sans-serif;
```
wenche, line 2: Don't set a font-family here, we already have the Equinor font

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+p {
+  margin: 0;
```
wenche, line 6: This one will improve when we use styled-components. But generally speaking, this is one of the problems with plain CSS as this rule will apply to each and every paragraph in the whole application.

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+  margin-bottom: 0.5rem;
+}
+
+.banner {
+  display: flex;
+  padding: 1rem;
+}
+
+.banner_container {
+  width: auto;
```
wenche, line 19: I don't think the width: auto is necessary here

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+  font-family: sans-serif;
```
wenche, line 20: Same, please don't override the font-family use the Typography component instead :)

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+  display: inline-block;
+  margin-right: 30px;
```
wenche, line 22: I don't think we need this margin as we already have a padding in the Banner component itself? And we should not mix px and rem

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+
+
+.banner--danger {
+  background-color: pink;
+}
+
+.banner--warning {
+  background-color:#FFE7D6;
+}
+.banner_icon   {
```
wenche, line 34: The EDS icon is not 2 rem, it is 24px. (With the default size that is). We shouldn't add an explicit width and height here as this causes spacing issues.

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+  width: 2rem;
+  height: 2rem;
+  margin-right: 0.5rem;
```
wenche, line 37: Redundant margin-right declaration, remove this

styles/banner.css
```diff
@@ -0,0 +1,40 @@
...
+  margin-right: 20px;
```
wenche, line 38: Ideally we should rather use `gap` here to get the spacing between the icon and the text. Not 100% sure about the support with flex-box for Safari tho.

Edit: support should be good enough https://caniuse.com/?search=gap

pages/checkout/terms.tsx
```diff
@@ -6,10 +6,16 @@ import { FormattedMessage, useIntl } from "react-intl"
...
 import styled from "styled-components"
 
+import { Banner } from "../../components/Banner"
 import { CheckoutWizard } from "../../components/CheckoutWizard/CheckoutWizard"
 import { Container } from "../../components/Container"
 import { Dispatch, RootState } from "../../store"
 
+const Title = styled.div`
+  display: flex;
+  column-gap: 0.5rem;
```
wenche, line 16: What do we need display and column gap for here? Should use the Typography component anyway :)

wenche (changes requested):

components/Banner/Banner.tsx
ShikhaMishra9: Need to have discussion on this one . I tried some things but still haven't got perfect results.

components/Banner/Banner.tsx
```diff
@@ -1,26 +1,33 @@
...
+const BannerIcon = styled.div`
+    padding-left: 5px;
+`
+
+export const Banner: FunctionComponent<Props> = ({ variant, children }) => (
+  <BannerContainer variant={variant}>
+    <BannerIcon>
+      <Icon color={variant === "danger" ? tokens.colors.interactive.danger__text.rgba : tokens.colors.interactive.warning__text.rgba} data={warning_filled} />
+    </BannerIcon>
+    <Typography>
```
wenche, line 29: <p> inside <p> is not allowed, so this Typography component should be replaced with a more generic div

pages/cart.tsx
```diff
@@ -82,50 +85,51 @@ const CartView : NextPage = () => {
...
+                  </Card>
 
-        <Banner variant="warning">
-          {intl.formatMessage({ id: "cart.banner.warning" })}
-        </Banner>
+                </CartItem>
+              ))}
+            </CartItems>
+          )}
+        <BannerContainer>
```
wenche, line 115: As discussed, it's better to remove this container and add the space to the banner component itself

components/Banner/Banner.tsx
```diff
@@ -1,26 +1,33 @@
...
-      <div className="banner_container">{children}</div>
-    </div>
-  )
-}
+const BannerIcon = styled.div`
+    padding-left: 5px;
+`
+
+export const Banner: FunctionComponent<Props> = ({ variant, children }) => (
+  <BannerContainer variant={variant}>
```
wenche, line 25: The banner container still needs some padding, and that one was removed by an accident

pages/checkout/terms.tsx
```diff
@@ -79,9 +74,9 @@ const CheckoutTermsView: NextPage = () => {
         {/* TODO: Add banner */}
         <BannerContainer>
           <Banner variant="danger">
-            <Title>{intl.formatMessage({ id: "terms.banner.danger.heading1" })}</Title>
+            <Typography variant="h4">{intl.formatMessage({ id: "terms.banner.danger.heading1" })}</Typography>
```
wenche, line 77: Use h2 level here and style the eds h2 according to our design :) And we probably need some space for the p as well.

But as discussed, this one might change if we are only doing a single checkout, so it's not 100% decided what the content in here should be. We will probably at some point need a more generic text content component that can be used in a map function for all the terms for all the data products coming from the API

wenche (changes requested):

components/Banner/Banner.tsx
wenche: We have discussed this and action points
- Remove the magic whitespace because of the inline element by adding `line-height: 0` to the wrapping icon div
- Add a flex-box and an align-items: center to align the text
- Test, as we just tested this quick and dirty 😅

wenche (changes requested):
One last thing 😅

wenche (approved):

PR #118, opened 2022-08-04
Update types for React 18

### LINKED ISSUE #117: Ensure that the code base is up to date with React 18
In order to avoid technical debt, we should double check that the code base is up to date with React 18. We have noticed at least one deprecated type warning for the `VoidFunctionComponent` type in the _document.tsx file and perhaps there is other warnings as well.

wenche:
resolves #117


nilsml (approved):
Looks good, but I believe CI does not agree?!

wenche:
> Looks good, but I believe CI does not agree?!

Do you think that is still the case?

nilsml:
> > Looks good, but I believe CI does not agree?!
> 
> Do you think that is still the case?

It doesn't matter what I think. The fact is that it is no longer an issue on CI 😄

PR #131, opened 2022-08-10
Removed extra a reference from button tag

ShikhaMishra9:


components/NavBar/NavBar.tsx
```diff
@@ -125,7 +125,7 @@ export const NavBar = () => {
         <HeaderContentContainer>
           <LogoContainer>
             <Link href="/">
-              <Button as="a" variant="ghost" color="secondary">
```
aronhoyer, line 128: what prompted this change?

ShikhaMishra9: It was earlier before  we updated the approach to use next link component rather than ours  . Updated the PR with link component from next now so this change not required.

wenche (approved):
LGTM 👍

PR #155, opened 2022-08-16
EDS Card component updated

ShikhaMishra9:


ShikhaMishra9:
Have added changes for Search.tsx from PR#154. Will be removed once that PR is merged

pages/index.tsx
```diff
@@ -71,17 +71,13 @@ const GridContainer = styled.div`
...
-`
-
 const AssetCardTitle = styled(Card.HeaderTitle)`
   font-weight: ${tokens.typography.paragraph.body_short_bold.fontWeight};
   color: ${tokens.colors.text.static_icons__default.hex};
   margin: 0;
   display: -webkit-box;
   -webkit-box-orient: vertical;
-  -webkit-line-clamp: 2;
+  -webkit-line-clamp: 1;
```
wenche, line 80: Only showing one line of the title is not much? 🤔

pages/index.tsx
```diff
@@ -151,17 +147,10 @@ const Frontpage: NextPage = () => {
...
-                href={{
-                  pathname: "/assets/[id]",
-                  query: { id: product.id },
-                }}
-                title={product.name}
-              >
+            <Link key={product.id} href={{ pathname: "/assets/[id]", query: { id: product.id } }} title={product.name}>
+              <Card elevation="raised">
                 <Card.Header>
                   <AssetCardTitle as="p">{product.name}</AssetCardTitle>
```
wenche, line 155: IMHO this would be improved by using a proper heading instead of `p`

pages/search.tsx
```diff
@@ -173,35 +183,48 @@ const Search: NextPage = () => {
...
+
+                    <Button variant="ghost_icon" color="secondary">
+                      <Icon data={list} />
+                    </Button>
+
+                    <Button variant="ghost_icon" color="secondary">
+                      <Icon data={gridOn} />
+                    </Button>
+                  </ViewModeActionsContainer>
+                )}
```
wenche, line 198: Is something wrong with the eslint setup somewhere since we get these kind of spacing issues? 🤔

pages/search.tsx
```diff
@@ -173,35 +183,48 @@ const Search: NextPage = () => {
...
-                      meta={[
-                        { label: intl.formatMessage({ id: "search.lastUpdated" }), value: Intl.DateTimeFormat("nb").format(new Date(resource.lastModifiedOn)) },
-                      ]}
-                    />
+                    <SearchResultItem key={resource.id}>
+                      <Link href={{ pathname: "/assets/[id]", query: { id: resource.id } }} title={resource.name}>
+                        <Card elevation="raised">
+                          <CardHeader>
+                            <CardHeaderTitle>
+                              <Typography variant="h4" as="p">
```
wenche, line 210: Same here, I don't agree with the `p` 😄

wenche (changes requested):
Some comments. 

Seems like you have not updated the cart view?

pages/index.tsx
ShikhaMishra9: Updated using heading

pages/search.tsx
ShikhaMishra9: This one i am not quite sure why its happening.

pages/index.tsx
```diff
@@ -152,7 +152,11 @@ const Frontpage: NextPage = () => {
             <Link key={product.id} href={{ pathname: "/assets/[id]", query: { id: product.id } }} title={product.name}>
               <Card elevation="raised">
                 <Card.Header>
-                  <AssetCardTitle as="p">{product.name}</AssetCardTitle>
+                  <AssetCardTitle>
+                    <Typography variant="h4">
```
wenche, line 156: For the front page, I'm adding a h1 for the new title in my PR and changing the "Product" header to h2. So correct level for the card title will then be `h3`

h1 -> h2 -> h3

pages/search.tsx
```diff
@@ -207,7 +207,7 @@ const Search: NextPage = () => {
                         <Card elevation="raised">
                           <CardHeader>
                             <CardHeaderTitle>
-                              <Typography variant="h4" as="p">
+                              <Typography variant="h4">
```
wenche, line 210: I think we are safe to assume that we will at some point have a h1 as a general title for the search here? 🤔  So in that case, the correct header level would be `h2`

That being said, EDS doesn't separate design and semantic level in a good way for the Typography component. For the envis project, we have wrapped the Typography component to get by some of the shortcomings. But I think you can do `<Typography variant="h4" as="h2" />` to make it look better but still use the correct level 😄

wenche (changes requested):

wenche (approved):

PR #157, opened 2022-08-16
Hiding See more from front page

ShikhaMishra9:


wenche (approved):
LGTM

aronhoyer:
should we maybe remove it altogether? i don't want us to end up with commented out code all over the place :grimacing:

ShikhaMishra9:
If there is no plan to have this feature added in front page then I am deleting it .

aronhoyer:
> If there is no plan to have this feature added in front page then I am deleting it .

i wouldn't say it's probable that it gets re-introduced, but there is of course a chance that it might be. but this is such a small thing to implement, so i don't see any value in keeping it.

PR #160, opened 2022-08-17
Improve the front page

### LINKED ISSUE #158: Improve the frontpage
*As a user I want to have a nice and good looking front page that explains the purpose of the Market place to make it easy to use and understand*

We do have some [nice suggestions for a front page in Figma](https://www.figma.com/file/54wzCBdXkEbWsAGXOVWLd6/Data-Marketplace?node-id=40%3A1576) including some illustrations. Is this something to implement? A: This is Petter's design, but we agreed to make something along these lines, where the text and/or illustration might be updated.

## Acceptance criteria
1. The front page has a title
1. The front page has a description that explains what data marketplace is
2. The front page has an illustration to set the mood

1. [Step 1]
1. [Step 2]

wenche:
Resolves #158


pages/_app.tsx
```diff
@@ -56,6 +57,9 @@ const App = ({ Component, pageProps }: AppProps) => (
   <MsalProvider instance={msalInstance}>
     <IntlProvider locale="en" defaultLocale="en" messages={englishTexts}>
       <Provider store={store}>
+        <Head>
+          <meta name="viewport" content="width=device-width, initial-scale=1" />
```
aronhoyer, line 61: doesn't nextjs add this out of the box?

pages/index.tsx
```diff
@@ -85,6 +86,35 @@ const AssetCardTitle = styled(Card.HeaderTitle)`
   overflow: hidden;
 `
 
+const Hero = styled.section`
```
aronhoyer, line 89: is there a reason you elected not to extend the `Section` styled component?

i feel like the search bar (which i believe we've decided to remove 🫣) is too close to the hero. `Section` has some nice margins that adds nice spacing between each section.

pages/index.tsx
```diff
@@ -85,6 +86,35 @@ const AssetCardTitle = styled(Card.HeaderTitle)`
...
+  }
+`
+
+const HeroContent = styled.div`
+  width: clamp(25ch, 60%, 600px);
+  z-index: 1;
+  align-self: start; 
+  background-color: rgba(255, 255, 255, 0.85);
+  border-radius: ${tokens.shape.corners.borderRadius};
+  padding: 0.5rem 0.5rem 0.5rem 0;
```
aronhoyer, line 106: smort! maybe the background-color should be solid white? and i agree (😏), maybe 0.5rem is a bit too sparse. we could probably increase it to 1 or even 1.5 rem

pages/index.tsx
```diff
@@ -131,7 +161,18 @@ const Frontpage: NextPage = () => {
           {documentTitle}
         </title>
       </Head>
+      <Hero>
+
```
aronhoyer, line 165: superfluous line feed? 🤔

aronhoyer (changes requested):

pages/_app.tsx
wenche: They didn't use to at least, and things suddenly started to look better when I added this 😅

pages/index.tsx
wenche: Haha, no. I actually on purpose didn't touch the the spacing and stuff with the search bar since we talked about to remove it. Should we just remove it, and fix the spacing?

pages/index.tsx
wenche: Dunno? It's a matter of taste I guess! But the contrast should be good enough with a semitransparent background as I double-checked that. Let's try 1rem

pages/index.tsx
aronhoyer: i don't think we should remove the navbar. even temporarily. i guess we could fix the spacing once we've replaced navbar with the eds topbar :+1:

wenche: Didn't talk about te navbar, just the search input field at the middle of the page

pages/index.tsx
wenche: Nah, didn't look too good with 1rem at least with the current text content. We could fine tune when we have the final text 👍  EDS does support mod 4 spacings, so it's not technically wrong

aronhoyer (approved):
lgtm :+1:

PR #164, opened 2022-08-19
Updated Top Bar

ShikhaMishra9:


ShikhaMishra9:
Moved search bar in center ....if its not ok will change

aronhoyer (changes requested):
added comment about commented out code.

looks good overall though.

components/NavBar/NavBar.tsx
```diff
@@ -105,58 +94,48 @@ export const NavBar = () => {
...
-
-          <SearchForm onSubmit={onSearchSubmit}>
-            <Search aria-label="sitewide" id="search-normal" placeholder={intl.formatMessage({ id: "navbar.placeholderSearch" })} onChange={onSearchChange} value={searchQuery} />
-          </SearchForm>
-        </HeaderContentContainer>
-      </Header>
-    </>
+  */}
+
+      <ActionsContainer>
```
wenche, line 100: According to the [EDS docs](https://eds-storybook-react.azurewebsites.net/?path=/docs/navigation-topbar--introduction#with-search-and-icons), the search field should be in the CustomContent and not the Actions container

wenche (changes requested):

wenche:
And also according to the docs: The Top bar should not scroll away :)

aronhoyer:
> And also according to the docs: The Top bar should not scroll away :)

is sticky not the default behaviour for topbar?

wenche:
> > And also according to the docs: The Top bar should not scroll away :)
> 
> is sticky not the default behaviour for topbar?

It should be, I think something is wrong in our code

wenche:
> > > And also according to the docs: The Top bar should not scroll away :)
> > 
> > 
> > is sticky not the default behaviour for topbar?
> 
> It should be, I think something is wrong in our code

Yes, our code does not actually use the Topbar component, it only uses the Action container. We need to use the component as it's done in the EDS documentatino @ShikhaMishra9

ShikhaMishra9:
> > > > And also according to the docs: The Top bar should not scroll away :)
> > > 
> > > 
> > > is sticky not the default behaviour for topbar?
> > 
> > 
> > It should be, I think something is wrong in our code
> 
> Yes, our code does not actually use the Topbar component, it only uses the Action container. We need to use the component as it's done in the EDS documentatino @ShikhaMishra9


Hi yes... It was giving some border . I have updated code with TopBar again .

components/NavBar/NavBar.tsx
```diff
@@ -61,22 +39,36 @@ const HeaderContentContainer = styled(Container)`
...
 
 const ActionsContainer = styled(TopBar.Actions)`
+  width: 100%;
+  background-color: ${tokens.colors.ui.background__default.hex};
   display: flex;
+  justify-content: flex-end;
+
+`
+
+const SearchContainer = styled(TopBar.CustomContent)`
```
wenche, line 60: Generally speaking, I think we are overdoing the styling here. We shouldn't need so much custom styles on the top of EDS. I think these custom styles can be removed all together and the same with the search form styles. If I do that, it will look like this and that's more aligned with EDS and a lot less styles
Skjermbilde 2022-08-19 kl  13 09 52

components/NavBar/NavBar.tsx
```diff
@@ -61,22 +39,36 @@ const HeaderContentContainer = styled(Container)`
     margin-right: 5rem;
```
wenche, line 39: I don't think this is needed? Isn't it just one child here anyway

components/NavBar/NavBar.tsx
```diff
@@ -61,22 +39,36 @@ const HeaderContentContainer = styled(Container)`
...
   }
 `
+const TopBarContainer = styled(TopBar)`
+  width: 100%;
+  padding-bottom: 4rem;
```
wenche, line 44: Why do we need a padding here, the height are supposed to be 64px according to EDS, now it's 74. Remove this padding 👍  If we need it for something, than something else is wrong

components/NavBar/NavBar.tsx
```diff
@@ -61,22 +39,36 @@ const HeaderContentContainer = styled(Container)`
...
+  padding-bottom: 4rem;  
 
+`
 const LogoContainer = styled.div`
   width: fit-content;
   white-space: nowrap;
 `
 
 const ActionsContainer = styled(TopBar.Actions)`
+  width: 100%;
```
wenche, line 53: Same here, are you sure that we need all these custom styles or are they just something that was needed in the past?

components/NavBar/NavBar.tsx
```diff
@@ -61,22 +39,36 @@ const HeaderContentContainer = styled(Container)`
...
+  display: flex;
+  justify-content: flex-end;
 
-  > *:not(:last-child) {
-    margin-right: 0.5rem;
-  }
 `
 
 const SearchForm = styled.form`
   width: 100%;
```
wenche, line 69: Same here, this styles are probably not needed

components/NavBar/NavBar.tsx
```diff
@@ -105,58 +97,29 @@ export const NavBar = () => {
...
-                  <Icon data={comment_discussion}
-                   title={intl.formatMessage({ id: "navbar.community" })} />
-                  <FormattedMessage id="navbar.community" />
-                </Button>
-              </Link>
-              */}
-            </ActionsContainer>
-          </nav>
-
+    <HeaderContentContainer>
```
wenche, line 100: You need to have the content container inside of the Topbar, since the Topbar is not sticky with respect to the HeaderContentContainer (the parent element). Since the HeaderContentContainer is the same height as the Top bar itself, you will not have the sticky effect. Try to swap them around and see if that doesn't help you with the sticky feature :)

wenche (changes requested):
As mentioned, I would recommend you to remove a lot of the custom CSS rules including all the flex-boxes. It look to me like the problem here is the good old bandaids on top of bandaids issue. 😅  (You can read about it [here](https://www.kevinpowell.co/article/dealing-with-the-%E2%80%9Cwebiness%E2%80%9D-of-css/) if you don't understand what I mean.)

If I remove a lot of the styles when I test locally, the Top bar looks a lot better. 👍  No wonder that it is like this, since it's been a lot back and forth, but take a couple of steps back and see if that makes things better.

I would also suggest to add the container that adds margin inside the component since the topbar bottom border now doesn't go all the way to the edges
Skjermbilde 2022-08-19 kl  13 28 09

wenche:
And one more thing btw: We might want to check all the pages in the application with the new top bar. I just noticed that the Cart page's title lacks some top spacing (we no longer need the Cart page, but might be an issue oter places as well)  because we removed the bottom 5rem margin of the top bar 👍  Might be something to add to the Container component or what do you think @aronhoyer ?

aronhoyer:
i'm thinking either we wrap `TopBar` and add it back, since sticky allows margin to affect layout. the other option is apply a `margin-top` to `section:first-child`. i don't think applying it to `Container` is the move, since container can be used in a context where top margin affects the layout in an unwanted way

components/NavBar/NavBar.tsx
aronhoyer: yeah the padding is a remnant from the two-row navbar. should just remove it :+1:

components/NavBar/NavBar.tsx
```diff
@@ -25,32 +20,15 @@ import { FormattedMessage, useIntl } from "react-intl"
...
-
-const Header = styled.header`
-  background-color: ${tokens.colors.ui.background__default.hex};
   width: 100%;
-  padding: 0.5rem 0;
-  position: sticky;
-  top: 0;
-  border-bottom: 2px solid ${tokens.colors.ui.background__light.hex};
-  margin-bottom: 5rem;
+  background-color: ${tokens.colors.ui.background__default.hex};
```
aronhoyer, line 31: we don't need a `background-color` any more

components/NavBar/NavBar.tsx
```diff
@@ -61,22 +39,36 @@ const HeaderContentContainer = styled(Container)`
...
+  width: 100%;
+  background-color: ${tokens.colors.ui.background__default.hex};
   display: flex;
+  justify-content: flex-end;
+
+`
+
+const SearchContainer = styled(TopBar.CustomContent)`
+  width: 80%;
+  background-color: ${tokens.colors.ui.background__default.hex};
```
aronhoyer, line 62: shouldn't need a `background-color` here either

aronhoyer (commented):
all in all there's a lot of unnecessary custom styling here.

i think an efficient way forward would be to remove all the custom styles, re-build the `NavBar` from EDS `TopBar` components and add necessary styling after.

if you want @ShikhaMishra9 we can have a walkthrough together

components/NavBar/NavBar.tsx
ShikhaMishra9: i have removed all styling and it works fine . Just looking for sticky thing as that one having some issue

aronhoyer: sticky is built-in to `TopBar`, so it should work if `TopBar` is the top most parent

components/NavBar/NavBar.tsx
```diff
@@ -24,59 +18,14 @@ import {
...
-const Header = styled.header`
-  background-color: ${tokens.colors.ui.background__default.hex};
-  width: 100%;
-  padding: 0.5rem 0;
-  position: sticky;
-  top: 0;
-  border-bottom: 2px solid ${tokens.colors.ui.background__light.hex};
+const EDSTopBar = styled(TopBar)`
   margin-bottom: 5rem;
+  margin-left: 8rem;
```
wenche, line 23: The Top bar would look better if it has 100% width, so we should remove these side margins.  With these margins added, the border doesn't look good and we get a strange behaviour while scrolling
Skjermbilde 2022-08-22 kl  08 26 45

wenche (changes requested):
Looks a lot better now! 👍  One last comment from my side

components/NavBar/NavBar.tsx
ShikhaMishra9: Updated

aronhoyer (approved):
nice work! :+1:

wenche (approved):

PR #173, opened 2022-08-23
Update communities filter with Checkbox

ShikhaMishra9:


wenche (changes requested):
Looks good 😄 

One improvement would be to add a little bit more semantic to the title and the checkbox list. If you look at this [EDS Checkbox example](https://eds-storybook-react.azurewebsites.net/?path=/docs/inputs-selection-controls-checkbox--introduction#grouped-checkboxes) you can see that a `fieldset` with a `legend` is used to group the checkboxes, and a `list` is used for the checkbox list. This will make things a lot easier to use with respect to a11y. 

A fieldset and legend have had a bad rumour as something that is incredibly hard to style, but I think this is less of a problem these days. You should of course aim to make it look like the Figma sketch.

ShikhaMishra9:
Updated with fieldset and list.

pages/search.tsx
```diff
@@ -55,20 +55,31 @@ const ViewModeActionsContainer = styled.div`
...
 `
 
 const SearchResultsList = styled(List)`
-  list-style: none;
+  list-style: none
+`
+
+const CommunityList = styled(List)`
+padding-inline-start: 0;
+font-weight: 500;
```
wenche, line 71: You can remove lines 71-73 because they have no effect due to the label styles. I guess you have tried to add the styles in the Figma sketches here? I don't think we should care about them as they are not according to EDS anyway. Using EDS compact as you already have done is more than enough 👍

wenche (changes requested):

pages/search.tsx
ShikhaMishra9: Removed font styling :D

wenche (approved):

PR #174, opened 2022-08-23
Improve the checkout flow

### LINKED ISSUE #151: Direct checkout
**As a User I want to checkout a data product directly from the asset page, so that I can more quickly get access to data**

Because of uncertainties around access request processing, we need to scrap batch checkout for the MVP. Instead the user should be able to start the checkout on a single data product, directly from the asset page.

## Sub tasks
- [ ] Checkout directly from the asset view by using a simple version with query params (PR #171)
- [ ] Revisit the checkout state management due to the single checkout
- [ ] Update the checkout wizard to use the new state management for the checkout flow

## Acceptance criteria
1. Re-label "add to cart" button
2. The user automatically gets directed to the checkout process
3. After successful redirect on last step

1. Click on a data product
2. Click "Get access" (might get a different label during design or implementation)
3. Complete the checkout process, which you should be directed to automatically

wenche:
We want to add a new and improved checkout flow. Partially resolves #151


components/CheckoutWizard/Stepper.tsx
```diff
@@ -0,0 +1,27 @@
...
+  grid-auto-flow: column;
+  grid-auto-columns: 1fr 1fr min-content;
+  padding: 0;
+`
+
+type Props = {
+  currentStep: number,
+}
+
+export const Stepper = ({ currentStep }: Props) => {
```
aronhoyer, line 18: i'd prefer we stick to using `FunctionComponent` as it sets the return type of the function as well

components/CheckoutWizard/types.ts
```diff
@@ -1,3 +1,3 @@
-export type CheckoutViewProps = {
-  assetId: string | null
+export type AssetIdProp = {
+  assetId: string | undefined
```
aronhoyer, line 2: when defining a type shape you can use `property?: type` to infer that it's either `type` or `undefined`

pages/checkout/access.tsx
```diff
@@ -119,7 +123,7 @@ const CheckoutAccessView = ({ assetId }: CheckoutViewProps) => {
 export const getServerSideProps: GetServerSideProps = async (context) => {
   const { id } = context.query
   // @TODO when we have server side token handle the case of no id or no data
-  return { props: { assetId: id || null } }
+  return { props: { assetId: id || undefined } }
```
aronhoyer, line 126: isn't `id || undefined` a little redundant?

pages/checkout/redirect.tsx
```diff
@@ -91,7 +91,7 @@ const CheckoutRedirectView = ({ assetId }: CheckoutViewProps) => {
 export const getServerSideProps: GetServerSideProps = async (context) => {
   const { id } = context.query
   // @TODO when we have server side token handle the case of no id or no data
-  return { props: { assetId: id || null } }
+  return { props: { assetId: id || undefined } }
```
aronhoyer, line 94: same here

pages/checkout/terms.tsx
```diff
@@ -122,7 +122,9 @@ const CheckoutTermsView = ({ assetId }: CheckoutViewProps) => {
 export const getServerSideProps: GetServerSideProps = async (context) => {
   const { id } = context.query
   // @TODO when we have server side token handle the case of no id or no data
-  return { props: { assetId: id || null } }
+  // We should query for terms and the asset title and evaluate how much this will
+  // affect TTFB
+  return { props: { assetId: id || undefined } }
```
aronhoyer, line 127: and here

pages/checkout/access.tsx
wenche: Hmm, I actually changed it back to `null` now since it was too easy to run into issues with SSR and serializing JSON. However I guess we will improve and change things here when we test SSR data fetching
Skjermbilde 2022-08-24 kl  14 38 48

aronhoyer: oooh true. when serializing JSON, `undefined` values are omitted

components/CheckoutWizard/types.ts
wenche: True that.

aronhoyer (approved):

PR #185, opened 2022-08-26
Using EDS Table for  for the responsibility tab

ShikhaMishra9:


components/AssetTabContent/ResponsibilitiesHolderList.tsx
```diff
@@ -36,19 +23,24 @@ export const ResponsibilitiesHolderList = ({
...
-          {holder.lastName}
-        </Typography>
-
-        <Typography variant="body_short" href={`mailto:${holder.email.toLowerCase()}`} link>{holder.email.toLowerCase()}</Typography>
-      </ResponsibilityRow>
-    ))}
-  </Container>
+  <Table>
+    <Table.Body>
+      <Typography variant="overline">{headline}</Typography>
```
wenche, line 28: I'm sorry, but this is not valid markup for a html table. In fact the HTML table element is very strict in which children it allows. You can for read more about html tables in[ the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table) and as you can see the only permitted childrens are table elements. 😄 

So what we need to do here is to add the paragraphs titles as `caption` (Data steward +++) and then add a `thead` elements with column headers. Like in the [EDS table example ](https://eds-storybook-react.azurewebsites.net/?path=/docs/data-display-table--introduction) I think it's fine to still use the overline style for the caption as the caption in this example is not looking too good IMHO 😅 

A table caption is also very useful for a11y. You can read more about tables and accessibility in [Difi's documentation](https://www.uutilsynet.no/veiledning/tabeller/225), it's a good read 👍

wenche (changes requested):

components/AssetTabContent/ResponsibilitiesHolderList.tsx
ShikhaMishra9: I referred to EDS table example here but little bit structure got mixed up . Have rectified that. Regarding table head we do not have any as in figma sketch. I have added an empty head for now but if you meant adding name and email id as head elements I can update that.

wenche: Yes, we should add name and email headers, the Figma sketches is not updated yet 👍

ShikhaMishra9: Added heading with Name and Email ID

wenche (approved):

PR #188, opened 2022-08-26
✨ Add cancel button for the checkout flow

### LINKED ISSUE #130: Implement Cancel button for checkout wizard
## Acceptance criteria
- When the user clicks on the `cancel` button in the checkout flow, he/she will be redirected back to the corresponding asset detail view

## Testing
- Click on an asset to get to the detail view
- Click on the `request access` button
- You should be taken to the checkout flow
- Click `cancel`
- You should be redirected to the asset detail view

wenche:
Resolve #130

### DIFF
--- a/components/CheckoutWizard/CancelButton.tsx
+++ b/components/CheckoutWizard/CancelButton.tsx
@@ -1,8 +1,7 @@
 import { Button } from "@equinor/eds-core-react"
+import Link from "next/link"
 import { FunctionComponent } from "react"
 import { useIntl } from "react-intl"
-
-import { Link } from "../Link"
 
 import type { AssetIdProp } from "./types"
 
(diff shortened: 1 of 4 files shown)


components/CheckoutWizard/CancelButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { Link } from "../Link"
+
+import type { AssetIdProp } from "./types"
+
+export const CancelButton: FunctionComponent<AssetIdProp> = ({ assetId }) => {
+  const intl = useIntl()
+
+  return (
+    <Link href={{ pathname: "/assets/[id]", query: { id: assetId } }}>
```
aronhoyer, line 13: renders `<a><a></a></a>`. our current `Link` component doesn't support `passHref` to children that are valid components. have created an issue to deprecate the `Link` component (#189)

i suggest we just use the next.js `Link` component going forward, as we've discussed earlier.

aronhoyer (changes requested):

components/CheckoutWizard/CancelButton.tsx
wenche: Djeeeez, right, I remember 😅  (Didn't test the markup 👀 )

aronhoyer (approved):
lgtm :+1:

PR #201, opened 2022-08-26
Clean up local storage

### LINKED ISSUE #176: Remove checkout content from local storage
We need to clean up in the local storage if a user aborts the checkout process and when they are finished to avoid stale data.

## Acceptance criteria
- Local storage is reset when a user navigates away from the checkout process
- Local storage is reset when the user clicks the cancel button
- Local storage is reset at the redirect step???

## To test
- Start the checkout process of an asset and complete the first two steps
- Verify that the data is stored in local storage
- Click cancel/navigate away -> the content is removed from local storage

wenche:
Resolves #176

### DIFF
--- a/components/CheckoutWizard/CheckoutWizard.tsx
+++ b/components/CheckoutWizard/CheckoutWizard.tsx
@@ -34,13 +34,19 @@ export const CheckoutWizard: FunctionComponent<Props> = ({ children, assetId }) 
   // const localCheckoutData = JSON.parse(window.localStorage.getItem("checkout_data") ?? "{}")
 
   useEffect(() => {
-    router.events.on("routeChangeComplete", () => {
+    const handleRouteChange = () => {
       const steps = ["terms", "access", "redirect"]
       const stepByPathName = steps.findIndex((step) => router.pathname.includes(step))
       // @TODO If a user manually refresh other steps than number 1, this logic
       // for current step will fail. See issue #83
       setCurrentStep(stepByPathName)
-    })
+    }
+
+    router.events.on("routeChangeComplete", handleRouteChange)
+
+    return () => {
+      router.events.off("routeChangeComplete", handleRouteChange)
+    }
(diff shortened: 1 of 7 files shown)


components/CheckoutWizard/CheckoutWizard.tsx
```diff
@@ -30,23 +32,41 @@ export const CheckoutWizard: FunctionComponent<Props> = ({ children, assetId })
...
+    router.events.on("routeChangeComplete", handleRouteChange)
+
+    return () => {
+      router.events.off("routeChangeComplete", handleRouteChange)
+    }
   }, [router])
 
+  useEffect(() => {
+    const handleRouteChange = (url: string) => {
+      const isInTheCheckoutFlow = steps.find((step) => url.includes(step))
```
aronhoyer, line 54: what if the user navigates to `/checkout/request-access` or `/checkout/access/some/nested/route`?

aronhoyer (commented):
looks good over all! just had a minor concern about the validation for whether or not a pathname is for a checkout step or not. to me it seems a little too accepting, but that might not be a big concern. let's discuss :+1:

components/CheckoutWizard/CheckoutWizard.tsx
wenche: You will not loose your data at least 😅  But yes, a little bit hard to define all the bits and pieces of this, I guess it might be better if the algorithm was 
`if has checkout as part of the path`
? Should we do it now or later

aronhoyer: yeah let's do that :+1:

wenche:
Ref. the other thing we talked about regarding security mechanism for locale storage and the asset id: I will create a new issue for that since it's a slightly bigger task than I initially thought

nilsml (approved):
IGILGTM

PR #205, opened 2022-08-29
Improve fetching of 6 most popular

### LINKED ISSUE #198: Get popular assets server side
**In order to improve the experience for the end user, we will get popular assets server side**

## Acceptance
- Popular assets are loaded server side to avoid loading spinners and content jump (CLS)

wenche:
Resolves #198 

The test with SSR is committed in `bf32988` but was reverted due to performance issues

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,11 +1,8 @@
 import { Card, Typography } from "@equinor/eds-core-react"
 import { tokens } from "@equinor/eds-tokens"
-import type { NextPage } from "next"
+import type { NextPage, GetServerSideProps } from "next"
+import { getToken } from "next-auth/jwt"
 import Head from "next/head"
-import {
-  useEffect,
-  useState,
-} from "react"
 import { FormattedMessage, useIntl } from "react-intl"
 import styled from "styled-components"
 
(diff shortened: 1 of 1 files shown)


pages/index.tsx
```diff
@@ -117,32 +117,33 @@ const Frontpage: NextPage = () => {
...
-
-            <GridContainer>
-              {popularDataProducts.length > 0 && popularDataProducts.map((product) => (
-                <Link key={product.id} href={{ pathname: "/assets/[id]", query: { id: product.id } }} title={product.name}>
-                  <CardGrid elevation="raised" onClick={() => {}}>
-                    <Card.Header>
-                      <AssetCardTitle>
-                        <Typography variant="h5" as="h2">
-                          {product.name}
+            {popularDataProducts.length > 0 ? (
```
aronhoyer, line 120: this makes the identical expression on line 122 redundant. suggest we just remove the check on line 122.

aronhoyer (changes requested):

pages/index.tsx
wenche: True that!

aronhoyer (approved):

PR #209, opened 2022-08-30
Get checkout asset data server side

### LINKED ISSUE #184: Get the asset name server side to avoid loading spinners on each checkout step
In order to improve the user experience we should pass the asset display name with the server side props. 

The easiest approach is probably to pass the name as a prop from the pages to the CheckoutWizard component.

## Acceptance
- Terms step will fetch the asset display name on the server
- Access step will fetch the asset display name on the server
- Redirect step will fetch the asset display name on the server

## Test
- There's no client side loading spinners for fetching the asset name

aronhoyer:
Resolves #184


wenche (approved):
LGTM 👍  Added #212

PR #210, opened 2022-08-30
Improve form validation

### LINKED ISSUE #175: Provide user feedback on invalid form input
_As a user I want the form to give me validation feedback so that it's easy to understand what I need to do_

For the checkout flow, we use the pattern of disabling the next button until the user has filled in the fields. This may be hard to understand for the user, and it also makes the form harder to use for screen readers as we don't  provide any information for the status of the field with aria attributes.

The description field also has a requirement of minimum 10 characters, but this is not communicated to the user. 

I would suggest to always have the buttons enabled, and rather use form validation techniques to give feedback and guide the users. This might be a little bit cumbersome to do manually, so we might want to consider to use react-hooks-forms. 

## Acceptance criteria - To be decided
1. [Criteria 1]
1. [Criteria 2]

1. [Step 1]
1. [Step 2]

wenche:
Resolves #175


pages/checkout/access.tsx
```diff
@@ -78,31 +84,24 @@ const CheckoutAccessView = ({ assetId }: AssetIdProp) => {
...
+                  multiline
+                  id="description"
+                  label={intl.formatMessage({ id: "checkout.access.descriptionInput.label" }, { maxLength: MAX_LENGTH })}
+                  placeholder={intl.formatMessage({ id: "checkout.access.descriptionInput.placeholder" })}
+                  onChange={onDescriptionChange}
+                  value={description}
+                  rows={4}
+                  variant={error ? "error" : "default"}
+                  maxLength={MAX_LENGTH}
+                  meta={`${description && description.length}`}
```
aronhoyer, line 97: renders `"undefined"` until the user starts typing

aronhoyer (changes requested):

pages/checkout/access.tsx
wenche: Outsmarted myself when I tried to fool EDS 😂

pages/checkout/access.tsx
```diff
@@ -20,34 +25,30 @@ const Ingress = styled(Typography).attrs(() => ({ variant: "ingress" }))`
...
   > *:not(:last-child) {
     margin-right: 1rem;
   }
 `
-
-const TextFieldContainer = styled.div`
-  margin-bottom: 1.5rem;
+const FakeHelperText = styled(Typography)`
+  /* Couldn't find this colour in the tokens, but it's the same as the helperText */  
+  color: rgb(111, 111, 111);
```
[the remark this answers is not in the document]
wenche: It is! How strange, tried to search for it in Zeroheight

nilsml (approved):
Looks good (except the Fake part)

PR #215, opened 2022-08-30
Improve the checkout design

### LINKED ISSUE #177: Update the design of the checkout flow
In order to make the checkout process with only one asset as easy as possible to understand we should update the code according to the latest [Figma](https://www.figma.com/file/54wzCBdXkEbWsAGXOVWLd6/Data-Marketplace?node-id=763%3A13434) sketches

## Acceptance criteria
- Content on terms is updated
- Content on access is updated
- Content on review is updated

wenche:
Resolves #177


locales/english.json
```diff
@@ -1,61 +1,41 @@
...
-  "spinner.loading": "Please hold\u2026",
-  "asset.addToCart": "Add to cart",
   "asset.getAccess": "Get access",
   "asset.description": "Description",
   "asset.purpose": "Purpose",
   "asset.timeliness": "Timeliness",
   "asset.overview": "Overview",
   "asset.responsibilites": "Responsibilites",
-  "terms.ingress": "<p>These terms and conditions create a contract between you as a data consumer and the data provider (the “Agreement”).</p><p>When you are provided access to this Data Product, you will have to ensure that the following criteria are met:</p>",
+  "terms.ingress": "<p>The agreement between you and the data provider is created by these terms and use.</p><p>Access to this Data Product requires you to meet the following criteria.</p>",
```
aronhoyer, line 26: it's either "terms **of use**" or "terms **and conditions**". collibra calls it "terms and conditions", so i think it might be best if we use that too.

locales/english.json
```diff
@@ -1,61 +1,41 @@
...
   "checkout.access.exampleBody": "As a [role] I want [description of what you want to do with the data product], so that [description of why you need the data product]",
-  "checkout.access.descriptionInput.label": "Description (max. {maxLength} characters)",
-  "checkout.access.descriptionInput.placeholder": "Please provide a description\u2026",
+  "checkout.access.descriptionInput.label": "Why do you need access? (max. {maxLength} characters)",
   "checkout.access.descriptionInput.errorMessage": "This field is required and you must add at least {minLength} characters",
-  "checkout.redirect.headline": "Redirecting to AccessIT",
-  "checkout.redirect.ingress": "You'll now get redirected to AccessIT to complete the request process.",
-  "checkout.redirect.body": "If you don't get redirected within 15 seconds, please click this link: <a>{link}</a>",
+  "checkout.redirect.headline": "We're redirecting you to AccessIT to complete the request.",
+  "checkout.redirect.body": "If you’re not redirected within 10 seconds, follow this link: <a>{link}</a>",
```
aronhoyer, line 33: incorrect apostrophe. we should probably use unicode escapes to be consistent on special characters

aronhoyer (changes requested):
looks good overall. just a few minor comments

locales/english.json
wenche: Discussed this back and forth with Trond yesterday, what we wanted was to use the same wording in the label for the checkbox and the ingress. Since the label used "Terms of use" we changed the ingress. Doesn't matter for me which term to use, but I would argue that it should be consistent.

wenche: Decided to use terms and conditions

aronhoyer (approved):

PR #222, opened 2022-09-01
Improved user feedback when we can't find terms and conditions

aronhoyer:
- Adds utility classes for three potential error "sources"
  - `ClientError` - any error caused by the user
  - `ExternalError` - any error caused by any third party (i.e. external APIs such as data sources)
- Introduces a schema for error codes: `EN000` where `N` is an enum denoting the source of the error, and the last three digits denote the type of error.
  - **0:** Errors caused by the user
  - **1:** Errors caused by an external resource
- Forwards the error code and message from `getServerSideProps` to the view
---
Resolves [AB#78408](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78408), resolves [AB#78740](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78740), resolves [AB#78741](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78741)

### DIFF
--- a/pages/checkout/terms.tsx
+++ b/pages/checkout/terms.tsx
@@ -80,10 +80,7 @@ const CheckoutTermsView: NextPage<Props> = ({ asset, error, rightsToUse }) => {
 
   useEffect(() => {
     if (asset) {
-      setCheckoutData({
-        ...checkoutData,
-        asset: { id: asset.id, name: asset.name },
-      })
+      setCheckoutData({ asset: { id: asset.id, name: asset.name } })
     }
   }, [asset])
 
@@ -167,7 +164,6 @@ const CheckoutTermsView: NextPage<Props> = ({ asset, error, rightsToUse }) => {
     asset,
     formError,
     hasAcceptedTerms,
-    intl,
     rightsToUse,
     checkoutData,
   ])
@@ -186,14 +182,8 @@ const CheckoutTermsView: NextPage<Props> = ({ asset, error, rightsToUse }) => {
 
 export const getServerSideProps: GetServerSideProps = async ({ req, query }) => {
   const { id } = query
-  // @TODO when we have server side token handle the case of no id or no data
-  // We should query for terms and the asset title and evaluate how much this will
-  // affect TTFB
-
   const defaultPageProps: Props = { asset: null }
-
   const token = await getToken({ req })
-
   const authorization = `Bearer ${token!.accessToken}`
 
   try {
--- a/pages/checkout/terms.tsx
+++ b/pages/checkout/terms.tsx
@@ -1,21 +1,33 @@
-import { Button, Checkbox, Typography } from "@equinor/eds-core-react"
+/* eslint-disable camelcase */
+import {
+  Banner,
+  Button,
+  Checkbox,
+  Icon,
+  Typography,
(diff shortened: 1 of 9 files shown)


nilsml (commented):
This approach will work, and has been used in similar variants for years. Personally I have never liked error handling with try/catch and have adopted a functional way of dealing with errors. I encourage you to read this article https://blog.logrocket.com/javascript-either-monad-error-handling/ to understand how we can handle errors without an Object Oriented approach.

Perhaps a discussion in the team about how we should deal with errors in general?

wenche:
- The article was a nice read, and it's refreshing to see the functional take on these kind of issues. I would really like to spend some time on this together with you @nilsml  sometime and utilise more of these techniques at some point in this project. For the time being I think this approach is OK tho.
- The useCallback wrapping makes the code a little bit harder to read, let's test it and see how it works out

pages/checkout/terms.tsx
```diff
@@ -49,83 +61,132 @@ const InfoBox = styled.div`
   margin-bottom: ${tokens.spacings.comfortable.medium}
 `
 
+const DataSourceErrorContainer = styled.div`
+  > *:not(:last-child) {
+    margin-bottom: ${tokens.spacings.comfortable.medium};
```
wenche, line 66: Side note: We might have reached the point where we could make life a little bit easer for ourselves if we added some css variables for the spacing tokens
`--spacing-medium: tokens.spacings.comfortable.medium`, `--spacing-md: tokens.spacings.comfortable.medium` `--spacing-8: tokens.spacings.comfortable.medium` or whatever floats your boat

pages/checkout/terms.tsx
```diff
@@ -49,83 +61,132 @@ const InfoBox = styled.div`
...
-  const onContinue = () => {
-    if (hasAcceptedTerms) {
-      setError(false)
-      router.push({
-        pathname: "/checkout/access",
-        query: { id: asset?.id },
+
+  useEffect(() => {
+    if (asset) {
+      setCheckoutData({ asset: { id: asset.id, name: asset.name } })
```
wenche, line 89: This will be updated twice in development, shouldn't we make it pure like ...asset 🤔

pages/checkout/terms.tsx
aronhoyer: they'd be added in `global.css`, right?

pages/checkout/terms.tsx
aronhoyer: that is true

aronhoyer: although, `asset` contains a lot more than just `name` and `id`, so it'd have to be `...(checkoutData.asset ?? {})`

pages/checkout/terms.tsx
wenche: Yes

pages/checkout/terms.tsx
wenche: Yup, whatever it takes to make it pure 😄

wenche (approved):
LGTM :thum

PR #224, opened 2022-09-02
Refresh access token five minutes before expiration

aronhoyer:
Resolves [AB#78644](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78644)


wenche (approved):
LGTM

PR #226, opened 2022-09-05
Properly set current checkout step if user refreshes

aronhoyer:
Resolves [AB#78531](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78531)


wenche (approved):
LGTM with router in the dependency array

PR #227, opened 2022-09-05
Prevent users from navigating to a checkout step without valid data

aronhoyer:
Resolves [AB#78528](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78528), resolves [AB#78530](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/78530)

### DIFF
--- a/components/CheckoutWizard/CheckoutWizard.tsx
+++ b/components/CheckoutWizard/CheckoutWizard.tsx
@@ -33,12 +33,24 @@ export const CheckoutWizard: FunctionComponent<Props> = ({ assetName, children }
   const [currentStep, setCurrentStep] = useState(0)
   const router = useRouter()
   const intl = useIntl()
-  const { removeItem } = useCheckoutData()
+  const { checkoutData, removeItem } = useCheckoutData()
 
   useEffect(() => {
     const stepByPathname = steps.findIndex((step) => router.pathname.includes(step))
+
+    if (stepByPathname > 0 && !(checkoutData as any)[steps[stepByPathname - 1]]) {
+      // this will recursively redirect until the user ends up at either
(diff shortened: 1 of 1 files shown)


components/CheckoutWizard/CheckoutWizard.tsx
```diff
@@ -33,12 +33,24 @@ export const CheckoutWizard: FunctionComponent<Props> = ({ assetName, children }
   const [currentStep, setCurrentStep] = useState(0)
   const router = useRouter()
   const intl = useIntl()
-  const { removeItem } = useCheckoutData()
+  const { checkoutData, removeItem } = useCheckoutData()
 
   useEffect(() => {
     const stepByPathname = steps.findIndex((step) => router.pathname.includes(step))
+
+    if (stepByPathname > 0 && !(checkoutData as any)[steps[stepByPathname - 1]]) {
```
wenche, line 41: Are we not able to type the checkoutdata?

wenche (approved):
Question about types, otherwise LGTM 👍 

The redirect behaviour was a little bit strange for the end user, but since it shouldn't happen unless you manipulate the url I think it's OK as it is a nice and clean solution of the problem

components/CheckoutWizard/CheckoutWizard.tsx
aronhoyer: we are, but `redirect` doesn't exist on `checkoutData` which typescript knows because of how `steps` is typed, so it gets mad when we use `steps` to index `checkoutData`

PR #230, opened 2022-09-05
🚸 Move the loading spinner to the result container

wenche:
This is an improvement to avoid to remove the title and the filters when we are updating the search result based on user input

### DIFF
--- a/pages/search.tsx
+++ b/pages/search.tsx
@@ -30,16 +30,12 @@ const SearchPageContainer = styled.div`
   display: grid;
   grid-template-columns: 15rem 1fr;
   grid-gap: ${tokens.spacings.comfortable.xx_large};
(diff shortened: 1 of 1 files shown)


pages/search.tsx
```diff
@@ -128,66 +124,66 @@ const Search: NextPage = () => {
...
+                      )
+                      : (
+                        <FormattedMessage
+                          id="search.results"
+                          values={{
+                            count: searchResults.length,
+                            searchTerm: (<b>{router.query.q}</b>),
+                          }}
+                        />
+                      ) }
```
aronhoyer, line 183: what is this formatting even 😂

aronhoyer (approved):
lgtm 👍

PR #232, opened 2022-09-06
Add pagination

wenche:
Use the total number of hits in the search result title, and adds a pagination component at the bottom of the search result list

### DIFF
--- a/pages/search.tsx
+++ b/pages/search.tsx
@@ -207,7 +207,7 @@ const Search: NextPage = () => {
                               <Typography variant="caption">
                                 {intl.formatMessage({ id: "search.lastUpdated" })}
                                 {": "}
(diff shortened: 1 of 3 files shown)


pages/search.tsx
```diff
@@ -114,6 +108,14 @@ const Search: NextPage = () => {
     })
   }
 
+  const onPaginationChange = (event: MouseEvent | KeyboardEvent | null, page: number) => {
+    const offset = page - 1
+    router.replace(
+      { query: { ...router.query, offset } },
+      { query: { ...router.query, offset } },
```
aronhoyer, line 115: what's going on here? 🤔

aronhoyer (changes requested):

pages/search.tsx
wenche: I want offset to be visible in the url? And would like shallow

aronhoyer (approved):

PR #233, opened 2022-09-06
Add missing <title>s

wenche:
Just a sausage while waiting


pages/checkout/access.tsx
```diff
@@ -101,6 +102,13 @@ const CheckoutAccessView: NextPage = () => {
 
   return (
     <>
+      <Head>
+        <title>
+          {intl.formatMessage({ id: "checkout.prefix.title" })}
+          -
```
aronhoyer, line 108: prefer unicode. i also see this snippet is repeated. maybe a move this to a utility function?

aronhoyer (changes requested):

pages/checkout/access.tsx
wenche: Sure 👍

aronhoyer (approved):

PR #235, opened 2022-09-07
Update dependencies

wenche:


wenche:
Are we really using ts-jest? Asking because it's commented out in jest.config.ts
`// preset: "ts-jest",`

aronhoyer:
> Are we really using ts-jest?

i don't think we are

wenche:
> > Are we really using ts-jest?
> 
> i don't think we are

Should we just remove it then? #bebrave

aronhoyer:
yeah, i checked out next.js' [jest + react-testing-library example repo](https://github.com/vercel/next.js/tree/canary/examples/with-jest)—which also has typescript—and i can't see ts-jest in the dependency list. so i say we just remove it

wenche:
Might be one of those things that comes out of the box with Next.js's jest support

aronhoyer (approved):

PR #238, opened 2022-09-08
Text and design adjustments for the frontpage

wenche:


locales/english.json
```diff
@@ -4,9 +4,10 @@
   "common.documentTitle": "Data Marketplace",
   "auth.loading": "Signing you in\u2026",
   "auth.prompt": "Please sign in",
-  "frontpage.hero.title": "Find and get access to the data you need, when you need it",
-  "frontpage.hero.ingress": "Data Marketplace makes it easy to search for and request access to the data you need to do your job. Venenatis ipsum aliquet nibh at. Risus sapien erat vitae amet leo lobortis maecenas.",
-  "frontpage.popularProductsHeader": "Popular",
+  "frontpage.hero.title": "Find the data you need",
+  "frontpage.hero.ingress": "Equinor Data Marketplace makes it easy to find and request access to the data you need.",
```
aronhoyer, line 8: didn't we agree to only call it "Data Marketplace"? 🤔

aronhoyer (changes requested):
pretty nifty solution on the topbar width! :+1:

missing some extra margin between the frontpage ingress and the banner below.
Screenshot 2022-09-09 at 10 35 24

maybe 1.5 or 2 rem.

see also my other comment

locales/english.json
aronhoyer: here's another suggestion to solve the orphan:
> Data Marketplace makes it easy for you to find and request access to the data you need, across Equinor.

wenche: Let's add that text, liked it :)

aronhoyer (approved):

PR #240, opened 2022-09-09
Refactor layout

wenche:
Part of  [AB#79312](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/79312)

### DIFF
--- a/components/AuthContainer/AuthContainer.tsx
+++ b/components/AuthContainer/AuthContainer.tsx
@@ -2,9 +2,18 @@ import { Typography } from "@equinor/eds-core-react"
 import { signIn, useSession } from "next-auth/react"
 import type { FunctionComponent, PropsWithChildren } from "react"
 import { useIntl } from "react-intl"
+import styled from "styled-components"
 
 import { Container } from "../Container"
-import { Page } from "../Page"
+
+const PageWrapper = styled.div`
+  min-height: 100vh;
+  display: flex;
+  flex-direction: column;
+  &:has(.background-highlight:last-child) {
+    background-color: var(--highlight-colour);
+  } 
+`
 
 export const AuthContainer: FunctionComponent<PropsWithChildren> = ({ children }) => {
   const { data, status } = useSession()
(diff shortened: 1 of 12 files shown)


wenche:
Some comments for AuthContainer:

- Not the ideal place for the PageWrapper, but just didn't bother to take the fight with Typescript 😅 
- Should we just remove the Container from the file and use a plain div or something instead? Would deprecate the Container component 100%

components/AuthContainer/AuthContainer.tsx
```diff
@@ -2,9 +2,18 @@ import { Typography } from "@equinor/eds-core-react"
 import { signIn, useSession } from "next-auth/react"
 import type { FunctionComponent, PropsWithChildren } from "react"
 import { useIntl } from "react-intl"
+import styled from "styled-components"
 
 import { Container } from "../Container"
-import { Page } from "../Page"
+
+const PageWrapper = styled.div`
```
aronhoyer, line 9: why is this here as opposed in `Page`? 🤔

pages/assets/[id].tsx
```diff
@@ -112,13 +112,12 @@ const AssetDetailView: NextPage<AssetDetailProps> = ({
   }
 
   return (
-    <>
+    <Page>
       <main>
         <Head>
           <title>{asset?.name}</title>
         </Head>
```
Comment (aronhoyer) [inline pages/assets/[id].tsx:119]: i think this is a good candidate to put into `<Page />`

aronhoyer (changes requested):

aronhoyer:
> * didn't bother to take the fight with Typescript

what was typescript complaining about this time?

pages/assets/[id].tsx
Comment (wenche) [reply pages/assets/[id].tsx:119]: That's a good idea :)

Comment (wenche) [reply pages/assets/[id].tsx:119]: > > didn't bother to take the fight with Typescript
> 
> what was typescript complaining about this time?

It didn't like it when I returned just the {children} and we don't need another wrapper

Comment (aronhoyer) [reply pages/assets/[id].tsx]: oh right, because a `FunctionComponent` has to return a `JSX.Element`

aronhoyer (approved):

PR #252, opened 2022-09-26
Update packages

wenche:
- Fix the issue with next-auth's dependency of `jose` that dependabot complains about
- Use the new slack icon from eds-icons
- Update the rest of the packages (minor and patch)


nilsml (approved):
LGTM

PR #251, opened 2022-09-23
Experiments with portable text

wenche:
Part of [AB#80910](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/80910)

PR #253, opened 2022-09-26
79767 add asset description to cards

aronhoyer:
There's currently an issue in Safari with rendering truncated text. This is likely due to the HTML structure in attributes provided by Collibra. Hopefully, we're able to resolve this with #251.

Resolves AB#79767

### DIFF
--- a/pages/api/popular.ts
+++ b/pages/api/popular.ts
@@ -1,45 +1,55 @@
 import type { NextApiHandler } from "next"
 import { getToken } from "next-auth/jwt"
+import xss from "xss"
 
-import { config } from "config"
-import { HttpClient } from "lib/HttpClient"
 import { HttpError } from "lib/HttpError"
+import { Asset } from "model/Asset"
+import { makeCollibraService } from "services"
+import { getAssetAttributes } from "services/collibra"
 
-type PopularAsset = Collibra.Asset & Pick<Collibra.NavigationStatistic, "numberOfViews">
+type PopularAsset = Asset & Pick<Collibra.NavigationStatistic, "numberOfViews">
 
 const getPopularAssets = async (
   data: PopularAsset[],
-  authorization: string,
+  r: ReturnType<typeof makeCollibraService>,
   limit: number,
   offset = 0,
 ): Promise<PopularAsset[]> => {
-  const mostViewedStats = await HttpClient.get<Collibra.PagedNavigationStatisticResponse>(`${config.COLLIBRA_BASE_URL}/navigation/most_viewed`, {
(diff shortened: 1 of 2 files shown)


aronhoyer:
@wenche should we wait for #251 to get merged before continuing with this branch?

wenche:
> @wenche should we wait for #251 to get merged before continuing with this branch?

Yes, let's wait

pages/api/popular.ts
```diff
@@ -1,45 +1,55 @@
...
+import { makeCollibraService } from "services"
+import { getAssetAttributes } from "services/collibra"
 
-type PopularAsset = Collibra.Asset & Pick<Collibra.NavigationStatistic, "numberOfViews">
+type PopularAsset = Asset & Pick<Collibra.NavigationStatistic, "numberOfViews">
 
 const getPopularAssets = async (
   data: PopularAsset[],
-  authorization: string,
+  r: ReturnType<typeof makeCollibraService>,
```
wenche, line 14: I find the `r` a little bit hard to read

pages/api/popular.ts
```diff
@@ -1,45 +1,55 @@
...
+  }))()
+
+  const collibraAssets = await Promise.all(
+    stats.map((stat) => r((client) => async (id: string) => client.get<Collibra.Asset>(`/assets/${id}`))(stat.assetId)),
   )
 
-  const result = [
+  const assets = collibraAssets
+    .map((res) => res.body)
+    .filter((asset) => asset.type.name.toLowerCase() === "data product")
```
wenche, line 31: Isn't this string "data product" hard coded several times in the code base at this point? Should we add a variable? 🤔

pages/api/popular.ts
aronhoyer: the type, or the parameter name?

pages/api/popular.ts
aronhoyer: two times, but yeah we should probably have a `isDataProductAssetType` utility function

wenche: Let's leave that as a future enhancement 👍

pages/api/popular.ts
aronhoyer: got you. i'll update it

wenche (approved):
LGTM

wenche:
#251 is finished soon, but perhaps we should do the truncate portable text issue in a separate issue? 🤔  I don't know, but this is also something that is needed for the search result cards and the portable text is still behind a feature flag. Isn't the Safari truncating bug visible in the search cards btw?

PR #256, opened 2022-09-29
Table component for Collibra data

wenche:
Solve #[AB#82489](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/82489)

Doesn't take formatted text inside the table cells into account (yet, will or will not be prioritised in another user story)

### DIFF
--- a/components/AssetTabContent/OverviewContent.tsx
+++ b/components/AssetTabContent/OverviewContent.tsx
@@ -7,7 +7,7 @@ import styled from "styled-components"
 
 import { AssetTabContentSectionContainer } from "./AssetTabContentSectionContainer"
 
-import { config } from "config"
+// import { config } from "config"
 import { defaultComponents } from "htmlParsing/portableText"
 
 const Overview = styled.div`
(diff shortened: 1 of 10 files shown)


nilsml (approved):
LGTM

components/AssetTabContent/OverviewContent.tsx
```diff
@@ -31,8 +31,9 @@ const OverviewSubTitle: FunctionComponent<PropsWithChildren> = ({ children }) =>
 )
 
 export const OverviewContent = ({ content }: Props) => {
-  const usePortableText = config.USE_PORTABLE_TEXT
-
+  const usePortableText = process.env.NEXT_PUBLIC_USE_PORTABLE_TEXT === "true"
```
aronhoyer, line 34: why was this removed from `config.ts`?

htmlParsing/rules.ts
```diff
@@ -0,0 +1,42 @@
+import { v4 as uuidv4 } from "uuid"
```
aronhoyer, line 1: node's built-in crypto package has a function to generate a v4 uuid. see https://nodejs.org/api/crypto.html#cryptorandomuuidoptions

components/AssetTabContent/OverviewContent.tsx
wenche: I left a comment about that one, it didn't update properly 🤷‍♀️  Was left with false server side and true client side, even after restarting the server ++

htmlParsing/rules.ts
wenche: Sweet, didn't know :)

components/AssetTabContent/OverviewContent.tsx
aronhoyer: that's peculiar 🤔

aronhoyer (approved):

PR #259, opened 2022-09-30
Prevent checkout of unapproved assets

aronhoyer:
Displays an error message if the asset the user tries to checkout
- an unapproved data product
- an asset that isn't a data product

Resolves [AB#81116](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/81116)

### DIFF
--- a/model/Asset.ts
+++ b/model/Asset.ts
@@ -2,6 +2,7 @@ type AssetInit = {
   id: string
   name: string
   description?: string
+  approved?: boolean
   domain?: string
   updateFrequency?: string
   tags?: DataMarketplace.Tag[]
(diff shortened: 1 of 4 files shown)


model/Asset.ts
```diff
@@ -2,6 +2,8 @@ type AssetInit = {
   id: string
   name: string
   description?: string
+  approved?: boolean
```
wenche, line 5: Is this not mandatory in collibra?

pages/checkout/terms.tsx
```diff
@@ -166,7 +168,7 @@ const CheckoutTermsView: NextPage<Props> = ({ asset, error, rightsToUse }) => {
             </>
           )
           ) }
-          {!asset && <NoAsset />}
+          {(!asset && !error) && <NoAsset />}
```
wenche, line 171: 😅

wenche (approved):
LGTM

model/Asset.ts
aronhoyer: the status field is, but the field will always have a value regardless of what the status is, so i made it mandatory :+1:

PR #260, opened 2022-09-30
Set up and configure Prettier

aronhoyer:


wenche:
For reference, here's the config we use for EDS and Energy Vision
```
trailingComma: all
tabWidth: 2
semi: false
singleQuote: true
arrowParens: always
endOfLine: lf
```

For Energy Vision, we also have
`printWidth: 120,`

I think some of them are the default values, like `arrowParens: always`

I would love to have Prettier enabled! As you say, eslint might be enough for 95% of the stuff but it has some shortcomings. I'm struggling with line lengths and manually having to add line breaks that in turn are overridden by other rules. 😅

aronhoyer:
rules like semi, and quote are already handled by eslint, but i do support 120 printWidth

wenche:
What do you think about this @ShikhaMishra9 ?

ShikhaMishra9:
I have used prettier earlier and i think that will be good to add :).

aronhoyer:
i've updated the config to use some of the rules you mentioned @wenche

wenche (approved):
LGTM 👍 

Let's take the discussion about trailing commas another time 🤺

aronhoyer:
you don't think `es5` is enough?

PR #261, opened 2022-09-30
fix: Only show approved data products in Popular assets

ShikhaMishra9:


pages/api/popular.ts
```diff
@@ -61,7 +67,23 @@ const PopularAssetsHandler: NextApiHandler = async (req, res) => {
...
+      query: { name: "data product" },
+    })
+
+    const dataProducts = await getPopularAssets(
+      [],
+      authString,
+      approvedStatusResp.body.id,
+      dataProductTypeIdRes.body.results[0]?.id,
+      limit,
+    )
```
aronhoyer, line 86: i think a possible improvement we can do to this, is if `dataProductTypeIdRes.body.results[0].id` is `undefined` (meaning there are no "data product" types [which should never really be the case, but just to guard against it]), we can respond to the client with a 500 status code or something.

next.js' api handlers use a node server framework called express.js, so sending a 500 status code would look like this: `res.status(500).end()`

aronhoyer (approved):

PR #262, opened 2022-09-30
Images for asset

wenche:
Resolve [AB#82558](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/82558)

### DIFF
--- a/htmlParsing/portableText/Image.tsx
+++ b/htmlParsing/portableText/Image.tsx
@@ -5,9 +5,7 @@ export const Image = ({ value }: any) => {
   const whereToPutThis = "https://equinor.collibra.com"
 
   return (
-    <>
-      <NextImage width={value.width} height={value.height} src={`${whereToPutThis}${value.src}`} alt={value.alt} />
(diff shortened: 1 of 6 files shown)


wenche:
I have done some investigation here, and the main issue is that we get `401` from Collibra when we try to show the image from the src attribute from the text string with the `img`  tag in our application. I did some experiments with the attachments/{:id} endpoint from the documentation, but the requested image is not available there (`404`) Might be because it is added inside of the rich text editor itself, and not as an attachment. The url for the image in the `src` attribute is for example` https://equinor.collibra.com/rest/latest/**attachment/download**/0d7d8f65-5908-4b85-ae38-6aefe30c6ee4`. 

We need to discuss this with the data catalog team.

wenche:
What should we do with this PR? It will not be completed in the short term. We do have the branch for this in the repository, and the most of the code is already in the code base since the parser for the img type is also needed to display the placeholder image. We also have an issue ab#833380 to look more into this at some point (it it's even possible)

It think we can just close til PR?

PR #268, opened 2022-10-05
Prevent unapproved assets from viewing

ShikhaMishra9:
[AB#81114](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/81114)

### DIFF
--- a/pages/api/popular.ts
+++ b/pages/api/popular.ts
@@ -77,6 +77,9 @@ const PopularAssetsHandler: NextApiHandler = async (req, res) => {
       query: { name: "data product" },
     })
 
+    if (dataProductTypeIdRes.body.results[0]?.id === "undefined") {
+      return res.status(500).end()
+    }
     const dataProducts = await getPopularAssets(
       [],
       authString,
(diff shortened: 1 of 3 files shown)


pages/api/popular.ts
```diff
@@ -77,6 +77,9 @@ const PopularAssetsHandler: NextApiHandler = async (req, res) => {
       query: { name: "data product" },
     })
 
+    if (dataProductTypeIdRes.body.results[0]?.id === "undefined") {
```
wenche, line 80: This is the one Petter suggested in the other PR?

wenche (approved):
LGTM 👍 

We should align with the case of no asset as well - we would probably like to return 404 if that's the case

pages/api/popular.ts
ShikhaMishra9: yes...should i put it in another PR

wenche: It's fine by me, just curious :)

PR #269, opened 2022-10-06
Check for empty search query

ShikhaMishra9:


ShikhaMishra9:
Yes for fixing the issue on front page .

PR #270, opened 2022-10-07
Avoid loading spinner at the search page

wenche:
Part of [AB#82304](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/82304)

### DIFF
--- a/components/Search/SearchResultStats.tsx
+++ b/components/Search/SearchResultStats.tsx
@@ -1,0 +1,33 @@
+import { FunctionComponent } from "react"
+import { FormattedMessage } from "react-intl"
+
+type Props = {
+  numberOfHits: number,
+  numberOfFilters: number,
+  query: string | string[]
+}
+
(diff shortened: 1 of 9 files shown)


components/Search/SearchResultStats.tsx
```diff
@@ -0,0 +1,33 @@
...
+  numberOfFilters: number,
+  query: string | string[]
+}
+
+export const SearchResultStats: FunctionComponent<Props> = (
+  { numberOfHits, numberOfFilters, query },
+) => {
+  if (numberOfHits === 0 && numberOfFilters > 0) {
+    return (
+      <FormattedMessage
```
nilsml, line 15: For cases like this we should have a conditional component. Multiple return statements like this makes for a more imperative code.

nilsml (approved):
For the most part it looks good!

PR #271, opened 2022-10-10
sm-AB#82371-404-page-not-found WIP

ShikhaMishra9:


pages/404.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <Hero>
+            <HeroContent>
+              <Typography variant="h1" style={{ marginBottom: "0.67em" }} bold>
+                {intl.formatMessage({ id: "notfound.hero.title" })}
+              </Typography>
+              <Typography style={{ marginBottom: tokens.spacings.comfortable.x_large }} variant="ingress">
+                <FormattedMessage
+                  id="notfound.hero.ingress"
+                  values={{
+                    a: FormattedLink,
```
wenche, line 77: I don't think this is the best way, even if I can see that it's done like this in the redirect url as well. :) I think it's a better approach the way it's done for the `p` in the terms.ingress in `terms.tsx` line ~134. The line height is also a little bit of compared to the rest of the text. We need to adjust that. Instead of using a link like this, we should use the Next link component with EDS typography instead. By doing that, we will use client side navigation with Next 👍  Also, I think it's better to use linkText or something like that instead of `here` in the translation file.

pages/404.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const HeroIllustration = styled(Illustration)`
+  width: clamp(500px, 40%, 500px);
+  justify-self: end;
+`
+
+const PageNotFound: NextPage = () => {
+  const intl = useIntl()
+
+  const FormattedLink = useCallback((chunks: ReactNode[]) => (
```
wenche, line 49: See comment below regarding the link

wenche (changes requested):
This page has a horisontal scroll because of the width of the SVG
`<svg width="1850" height="226" viewBox="0 0 350 226" fill="none" xmlns="http://www.w3.org/2000/svg">` Seems like it is still not the correct frame from Figma?

The 404 page should work even if the window is narrow :)

wenche:
Nice work Shikha :) The most important things with this task was probably
1) Be careful when you export the SVG from Figma to ensure that the correct frame is selected
2) Spread rest props and apply (or add className/style if more control is needed) to make it possible to add style to a component from outside using styled-components or similar

ShikhaMishra9:
> Nice work Shikha :) The most important things with this task was probably
> 
> 1. Be careful when you export the SVG from Figma to ensure that the correct frame is selected
> 2. Spread rest props and apply (or add className/style if more control is needed) to make it possible to add style to a component from outside using styled-components or similar

Yes I understood the importance of frame here  :)

PR #272, opened 2022-10-10
refactor: get terms related to data product

aronhoyer:
Resolves [AB#83757](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/83757)


aronhoyer:
i'm aware that this pr expands on already existing oop principals. maybe @nilsml and i could take some time and pair program some new models that follow functional paradigms more closely 😇

nilsml (commented):
Like you say it is following the oo principles. We need to make sure we don't go down this route for too long as refactoring oo code to fp is harder than solving it in a functional way in the first place.

wenche (commented):
What NM said. And this seems like a much more robust and better way to fetch the terms and conditions! 😅

aronhoyer:
does that block this pr though?

wenche:
> does that block this pr though?

No, let's do it!

aronhoyer:
ok but no-one's approved it 😅

wenche:
Ah sorry, thought Nils Magne did that :D

wenche (approved):

PR #273, opened 2022-10-10
Use env variable for auth scope

wenche:
Resolve [AB#83872](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/83872)


nilsml (approved):
Seems ok. Let's see if it works in prod

PR #274, opened 2022-10-17
Disable Collibra workflow while user testing

ShikhaMishra9:


.env.template
```diff
@@ -14,4 +14,5 @@ NEXTAUTH_SECRET=
 ANALYZE=true|false
 
 # === Dummy feature flags ===
-NEXT_PUBLIC_USE_PORTABLE_TEXT=true|false
\ No newline at end of file
+NEXT_PUBLIC_USE_PORTABLE_TEXT=true|false
+NEXT_PUBLIC_USE_COLLIBRA_TEST=true|false
```
aronhoyer, line 18: i would suggest a more descriptive name for this. think about what we're actually trying to complete with having this flag.

aronhoyer (changes requested):
generally, it seems like you've understood the idea of feature flagging with environment variables. however, the feature flag name doesn't fit with what it's flagging; `USE_COLLIBRA_TEST` isn't what we're toggling. rather we're toggling between whether we want to trigger a collibra `request access workflow`. furthermore, it doesn't seem that this will actually prevent the workflow to get triggered.

remember when we discussed this, that we concluded that in `pages/checkout/access.tsx`, there's an on click handler that sends a POST request to `/api/workflows` that we want to prevent from getting sent. it's that POST request that triggers the workflow.

additionally, in `config.ts` you've created an entry that'll evaluate to a boolean, but never use it. rather you're comparing `process.env` directly. you'd be better off using the entry you've created in `config.ts`

.env.template
ShikhaMishra9: Hi sure will change name ...Also i  thought wrt user test it need to get restricted . will add this condition at access.tsx

aronhoyer: yes, user test is the short term use case, but we generally want variable names want to describe what they're accomplishing in the program. so something like `PREVENT_WORKFLOW_TRIGGER` or `FULFILL_CHECKOUT`

ShikhaMishra9: Updated PR

pages/checkout/access.tsx
```diff
@@ -79,13 +69,16 @@ const CheckoutAccessView: NextPage = () => {
...
         setIsLoading(true)
-        await HttpClient.post<Collibra.WorkflowInstance>("/api/workflows", {
-          body: {
-            assetId: checkoutData.asset?.id,
-            termsAccepted: checkoutData.terms?.termsAccepted,
-            description: checkoutData.access?.description,
-          },
-        })
+
+        if (config.ACCESS_COLLIBRA_WORKFLOW) {
```
aronhoyer, line 73: after having tested this locally, i think we should reverse the logic here. with the currently proposed implementation, we have to opt in to trigger the collibra workflow. this means that this will change the behaviour of prod, which we don't want feature flags to do, unless toggled.

aronhoyer: additionally, the variable name should be changed to reflect that it's opt in. so instead of `ACCESS`, you could name it something like `BLOCK` or `PREVENT` 😄

aronhoyer (approved):
looks good! 👍

PR #277, opened 2022-10-25
Custom sign in page

wenche:
Resolve [AB#81151](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/81151)

[AB#81152](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/81152) and [AB#85594](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/85594)


components/AuthContainer/AuthContainer.tsx
```diff
@@ -1,30 +1,20 @@
...
 
   if (status !== "authenticated" || data?.error === "RefreshAccessTokenError") {
-    signIn("azure-ad")
-
-    return (
-      <Container>
-        <Typography>{intl.formatMessage({ id: "auth.prompt" })}</Typography>
-      </Container>
-    )
+    return <SignIn />
```
aronhoyer, line 17: i think we might want to do this server side and redirect to the route instead of just rendering the view

aronhoyer (approved):

middleware.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line no-restricted-exports, import/no-default-export
+export { default } from "next-auth/middleware"
+
+export const config = { matcher: ["/", "/assets/:path*", "/checkout/:path*", "/search"] }
```
aronhoyer, line 4: in a perfect world, we'd be able to tap into the middleware and if the user isn't authorized, redirect them to `/auth/signin?origin={route}`. do you think we'd be able to that at all?

wenche: I'm not sure if I see why you would prefer that pattern? If we don't want every page to be secured, we would need some white/black-listing anyway. If not (the way it actually is now) I would very much prefer that the [documented way of protecting all pages](https://next-auth.js.org/tutorials/securing-pages-and-api-routes#nextjs-middleware) just worked, but it doesn't 🤷‍♀️ 

But I guess it is possible by using the `unstable_getServerSession` to check for an active session in the middleware and redirect if not? 🤔

aronhoyer: does the current solution redirect the user to the page they were on before getting the sign in page?

aronhoyer: i guess it's not a huge deal if it doesn't

wenche: I does, due to the last commit. Men nekter egentlig å belite meg på at det skal være nødvendig 👀

aronhoyer:
are we planning to initialize the sign in flow when the user clicks "Create an account" a later pr in, or should we do it in this pr?

wenche:
> are we planning to initialize the sign in flow when the user clicks "Create an account" a later pr in, or should we do it in this pr?

ab#85593

aronhoyer:
so that means "later"?

wenche:
Yes, later, but probably the next one in the queue don't you think? Could also be something for @ShikhaMishra9  to look into

aronhoyer:
sounds good.

as a side note, i believe this pr also resolves AB#81152 and AB#85594

wenche:
> sounds good.
> 
> as a side note, i believe this pr also resolves [AB#81152](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/81152) and [AB#85594](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/85594)

Yes, I think you are right about that. At least for the time being :)

PR #286, opened 2022-11-03
🔧 Prefer const for eslint

nilsml:
I think we should make it difficult to use let and var

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,6 +42,7 @@
       "rules": {
         "@typescript-eslint/no-shadow": ["error"],
         "@typescript-eslint/no-unused-vars": ["error"],
+        "prefer-const": "warn",
         "no-shadow": "off",
         "no-undef": "off",
         "no-unused-vars": "off"


wenche (approved):
LGTM 👍

aronhoyer:
this rule is already set in `eslint-config-airbnb`

nilsml:
I thought so too, but I don't see any warnings when using let. Perhaps change to error instead of warning 🤔

aronhoyer:
eslint-config-airbnb has it set to error already (see https://github.com/airbnb/javascript/blob/366bfa66380c08304101c6add46355696e90b348/packages/eslint-config-airbnb-base/rules/es6.js#L115-L119)

wenche:
I get this in my IDE. I don't know if no-mutable-exports is supposed to be a substitute or improvement to prefer-const, but I guess this is good enough for us? Close this PR?

Skjermbilde 2022-11-29 kl  13 13 11

aronhoyer:
yeah, i think this pr is ripe for closing 👍

PR #288, opened 2022-11-14
refactor: Make NEXT_PUBLIC_COLLIBRA_BASE_URL non-public

aronhoyer:
Since `COLLIBRA_BASE_URL` is only used server side, there's no need for it to keep the `NEXT_PUBLIC_` prefix.

**IMPORTANT NOTE**
This change requires an update to GitHub secrets to build. Not sure whether we want to do this before or after this is merged.

Resolves [AB#88795](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/88795)

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,5 +1,5 @@
 # === Collibra ===
-NEXT_PUBLIC_COLLIBRA_BASE_URL=
+COLLIBRA_BASE_URL=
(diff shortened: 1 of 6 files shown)


wenche (approved):
LGTM 👍

PR #289, opened 2022-11-22
fix: Fix issue where access token didn't refresh

aronhoyer:
Fix issue where access token didn't refresh on first visit after expiration

Needs validation when refresh token expires, but should work in theory

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -8,16 +8,7 @@ type HttpClientConfig = {
   headers?: Record<string, any>
 }
 
-type HttpMethod = "GET"
-| "DELETE"
-| "HEAD"
-| "OPTIONS"
-| "POST"
-| "PUT"
-| "PATCH"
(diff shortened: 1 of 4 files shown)


lib/HttpClient.ts
```diff
@@ -8,16 +8,7 @@ type HttpClientConfig = {
...
-| "DELETE"
-| "HEAD"
-| "OPTIONS"
-| "POST"
-| "PUT"
-| "PATCH"
-| "PURGE"
-| "LINK"
-| "UNLINK"
+type HttpMethod = "GET" | "DELETE" | "HEAD" | "OPTIONS" | "POST" | "PUT" | "PATCH" | "PURGE" | "LINK" | "UNLINK"
```
aronhoyer, line 11: prettier PLEASE

wenche (approved):
LGTM 👍  Let's test and see how it works

PR #292, opened 2022-11-24
fix: Handling of expired refresh token

aronhoyer:
Resolves [AB#90339](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/90339)


wenche (approved):
LGTM 👍  interesting to see if that solves the problem!

PR #296, opened 2022-11-29
Update feedback mechanism to use a Teams channel instead of Slack

wenche:
EDS doesn't have the Teams icon in eds-icons

Resolves ab#88342


wenche:
We couldn't use Teams to create an open, public channel the way we wanted to

PR #300, opened 2022-12-13
Fix bug with search, pagination and browser back

wenche:
Resolve [AB#92269](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/92269)


wenche:
I think this rewrite will probably be a better approach. 🤔  The issue was caused by a race condition in the effect because the useEffect also ran when browser back was activated (because router is in the dependency array according to hooks best practice). That's also why it was not consistent.

I think it we do it like this instead, things are more clean and correct. What we want is just to add the active tab to the query param, and that will happen when the user changes tab. We don't set the overview tab as default in the url, by I see that as a good thing since it's a little bit confusing that the url changes without user interaction

aronhoyer (approved):
thank you for deleting `Link` 😅

PR #305, opened 2022-12-23
Upgrade packages + bonus bug fix with 404

wenche:


locales/english.json
```diff
@@ -21,7 +21,7 @@
   "frontpage.numberOfViews": "{numberOfViews} views",
   "notfound.code.number": "404",
   "notfound.hero.title": "Uh-oh This page doesn't exist",
-  "notfound.hero.ingress": "Seems like you've gotten lost. Don't worry, though. We'll help you find your way back. Just click <a>Link</a>.",
+  "notfound.hero.ingress": "Seems like you've gotten lost. Don't worry, though. We'll help you find your way back. <a>Take me back to the start page.</a>",
```
aronhoyer, line 24: we're mixing speaking **to** the user and **for** the user within the same paragraph here. I like the wording, but can we change it to "go back to the home page" or something?

aronhoyer (changes requested):

PR #312, opened 2023-01-02
🔥 Remove (hopefully) unused application config variable

wenche:
Remove an old application insight connection string from the config file.


aronhoyer (approved):
why is `package-lock.json` updated? 🕵️

wenche:
> why is `package-lock.json` updated? 🕵️

Wondered about the same thing, but I think it's because I ran a npm i due to all the other package upgrades on main

PR #314, opened 2023-01-05
feat: Hide more information button if excerpt is nullish

aronhoyer:
Resolves [AB#94185](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/94185)

### DIFF
--- a/components/AssetTabContent/OverviewContent.tsx
+++ b/components/AssetTabContent/OverviewContent.tsx
@@ -20,6 +20,10 @@ const Overview = styled.div`
   max-width: 70ch;
 `
 
+const MoreInfoContainer = styled.div<{ alone: boolean }>`
+  ${({ alone }) => alone && `margin-top: ${tokens.spacings.comfortable.medium};`}
+`
+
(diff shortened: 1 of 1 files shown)


components/AssetTabContent/OverviewContent.tsx
```diff
@@ -20,6 +20,14 @@ const Overview = styled.div`
   max-width: 70ch;
 `
 
+const MoreInfoContainer = styled.div<{ alone: boolean }>`
+  ${({ alone }) =>
```
wenche, line 24: Sounds so sad #4alone 😄

wenche: btw, I think you could just do it like this and avoid the css import
`  ${({ hasDropShadow }) => hasDropShadow && `box-shadow: var(--topBar-box-shadow);`}`

wenche (approved):
Might or might not be better to skip css, otherwise LGTM

PR #316, opened 2023-01-09
Fetch asset info using the adapter service

wenche:
Resolves [AB#94811](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/94811)


wenche:
I'm not sure if we need to handle the potential case of a missing AccessIT url value in the adapter service? Can we assume that if the authURLattr exists, then the value will exist as well? In other words that the emptyAttr conditional is enough?

Also, we should look into how we handle potential missing data and errors in the user interface - I don't think this is very robust or aligned across different pages

pages/checkout/redirect.tsx
```diff
@@ -113,29 +109,31 @@ export const getServerSideProps: GetServerSideProps = async ({ req, query }) =>
 
   const token = await getToken({ req })
 
-  if (typeof token?.accessToken !== "string") {
-    return { props: defaultPageProps }
+  if (!token || !token.accessToken || typeof token.accessToken !== "string") {
```
aronhoyer, line 112: this change seems unnecessary 🤔

wenche: Maybe? It was mainly done to align with the rest of the code base 😅  It was done like that in /assets/{id}, then I updated it in /terms and now here in /redirect. Should either keep it, or adjust it in the other files? I just want to do this in the same way 👍

aronhoyer:
> I'm not sure if we need to handle the potential case of a missing AccessIT url value in the adapter service?

probably a good idea to catch it, but ideally, it should be defined as long as the asset is approved. right?

pages/checkout/redirect.tsx
aronhoyer: i mean it all does the same thing, so it really doesn't matter beyond semantics. besides, comparing booleans is cheap. however, my two cents is that this is overly verbose, and `typeof undefined?.key === "type"` will evaluate to false

wenche: I think we should stick to one of them to avoid confusion at least, since the code is quite similar for the first part across the 3 pages. We can change all 3 to the "short" version?

wenche:
> > I'm not sure if we need to handle the potential case of a missing AccessIT url value in the adapter service?
> 
> probably a good idea to catch it, but ideally, it should be defined as long as the asset is approved. right?

Yes, that's my assumption as well.

pages/checkout/redirect.tsx
aronhoyer: i'm partial to optional chaining, but it's your call

aronhoyer (approved):

PR #317, opened 2023-01-13
✨ Fetch data server side for front page with adapter service

wenche:
I want to measure the performance of doing a server side fetch of popular data since that is more in line with how we do the rest. It's been a while since we tried this


wenche:
This still feels a little bit too slow, unfortunately. Getting good results with lighthouse, but the initial server response time is too high IMHO and this feels laggy (1.5-2 secs)

PR #321, opened 2023-01-16
Remove local models, and use the data marketplace model package instead

wenche:
Resolves [AB#96104](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/96104)


wenche:
I have deleted a lot of code 🤞  that I didn't remove _too_ much. I also fixed some eslint stuff, but the last on is a though one. We should look into this at some point, it's a weak part of the code

ShikhaMishra9 (approved):
Looks good :)

PR #329, opened 2023-01-24
⚗ Add excerpt and description as snippets

wenche:
Resolves [AB#97096](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/97096)

We are still in the experimental phase where we explore the opportunities we have

### DIFF
--- a/components/ImprovedSearch/Hit.tsx
+++ b/components/ImprovedSearch/Hit.tsx
@@ -2,12 +2,19 @@ import { Card, Typography, Chip } from "@equinor/eds-core-react"
 import { tokens } from "@equinor/eds-tokens"
 import type { Hit as AlgoliaHit } from "instantsearch.js"
 import NextLink from "next/link"
-import { Highlight } from "react-instantsearch-hooks-web"
+import { Highlight, Snippet } from "react-instantsearch-hooks-web"
 import styled from "styled-components"
 // import type { SendEventForHits } from "instantsearch.js/es/lib/utils"
 const { Header: CardHeader, HeaderTitle: CardHeaderTitle } = Card
 
 const StyledHightLight = styled(Highlight)`
+  & .highlighted {
+    background-color: ${tokens.colors.interactive.primary__selected_highlight.hsla};
+  }
+`
+
+const StyledSnippet = styled(Snippet)`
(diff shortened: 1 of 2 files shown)


components/ImprovedSearch/Hit.tsx
```diff
@@ -13,6 +13,13 @@ const StyledHightLight = styled(Highlight)`
   }
 `
 
+const StyledSnippet = styled(Snippet)`
+  color: ${tokens.colors.text.static_icons__default.hsla};
+  & .highlighted {
+    background-color: ${tokens.colors.interactive.primary__selected_highlight.hsla};
```
aronhoyer, line 19: note for future design: this definitely needs higher contrast.

pages/search-beta.tsx
```diff
@@ -53,15 +53,15 @@ const Search = ({ serverState, isServerRendered, serverUrl }: Props) => (
...
+        indexName="Data_Set"
         routing={{
           router: history({
             /*  @ts-ignore */
             getLocation: () => (typeof window === "undefined" ? new URL(serverUrl) : window.location),
           }),
         }}
       >
-        <Configure hitsPerPage={50} snippetEllipsisText="..." />
+        <Configure hitsPerPage={50} snippetEllipsisText="..." attributesToSnippet={["excerpt", "description"]} />
```
aronhoyer, line 64: would there be a way, do you think, for us abstract the `attributesToSnippet`? in case the model changes, so we don't have hard coded values

aronhoyer (commented):
looks good to merge.
just a note and a curiosity from my end

pages/search-beta.tsx
wenche: Good question 👍  When we have created the search indexing service, the `attributesToSnippet` will be added there. The Configure is just a way to do things until we get there (might be that we need it for hitsPerPage, I don't remember 100%)

[See here in Envis for an example](https://github.com/equinor/energyvision/blob/main/search/IndexSanityContent/topic/algolia.ts)

aronhoyer (approved):
forgot to approve 😅

PR #334, opened 2023-01-30
Fix access token refresh

aronhoyer:
This pull request aims to resolve the issues we've been facing with users receiving 401s from the first API requests performed when the app is opened after a considerable amount of time of inactivity.

The solution isn't the most elegant, but it boils down to running `getToken` again if the API call fails with a 401.

[SWR](https://swr.vercel.app) should allow us to make retrying API requests easier, but out of scope for this pull request, as it'd require a larger rewrite.

Solves [AB#97995](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/97995)

### DIFF
--- a/collibra.d.ts
+++ b/collibra.d.ts
@@ -1,7 +1,521 @@
 declare namespace Collibra {
+  type ResourceType =
+    | "View"
+    | "Asset"
+    | "Community"
+    | "Domain"
+    | "AssetType"
+    | "DomainType"
+    | "Status"
+    | "User"
+    | "ClassificationMatch"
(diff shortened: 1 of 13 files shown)


nilsml (approved):
Had a good walk-through with @aronhoyer, and I understand the flow now.
I don't understand what next-auth is doing under the hood though, and find it strange that we need to handle this issue in such a way. I think we should merge this in and monitor. Also make an issue or note to look for a better solution some time down the road with new versions of the auth library.

PR #335, opened 2023-01-30
Add filter for improved search

wenche:
Resolves [AB#96929](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/96929)


wenche:
> Looks good ... Saw that you have used EDS component as well in RefinementList

The provider? Yes, because of the compact mode. Just pasted it from the original search :)

PR #337, opened 2023-01-31
Adding paginattion component for search page

ShikhaMishra9:


aronhoyer:
doesn't the [eds pagination component](https://eds-storybook-react.azurewebsites.net/?path=/docs/navigation-pagination--introduction) work with algolia?

wenche:
> doesn't the [eds pagination component](https://eds-storybook-react.azurewebsites.net/?path=/docs/navigation-pagination--introduction) work with algolia?

No, see the comment on the issue at the board :)

components/ImprovedSearch/Button.tsx
```diff
@@ -0,0 +1,20 @@
+import { forwardRef } from 'react'
+import styled from 'styled-components'
+import { Button as EdsButton, ButtonProps as EdsButtonProps } from '@equinor/eds-core-react'
+
+export type ButtonProps = EdsButtonProps
+
+const StyledEdsButton = styled(EdsButton)`
+  --eds_button__font_size: var(--typeScale-05);
```
wenche, line 8: We don't have these CSS variables in our code base, so this doesn't actually do anything 🤷‍♀️  I think you should reconsider if we need this button as a separate component at all

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { chevron_left, chevron_right, first_page, last_page } from '@equinor/eds-icons'
+import { useContext, useEffect, useRef } from 'react'
+import { usePagination, UsePaginationProps } from 'react-instantsearch-hooks-web'
+import styled from 'styled-components'
+import { usePrefersReducedMotion } from './ReducedMotion'
+import { PaginationContext } from './PaginationContext'
+import { PaginationItem } from './PaginationItem'
+
+
+const PaginationList = styled.ul`
```
wenche, line 11: You should use the EDS list here, like we do for the filters and the hits list

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+  margin: 0;
+  display: grid;
+  grid-gap: 8px;
+  grid-auto-columns: min-content;
+  grid-auto-flow: column;
+`
+
+export type PaginationProps = {
+  hitsPerPage?: number
+  inverted?: boolean
```
wenche, line 23: We don't need the inverted option. It's used by Envis because we use this component both on white and dark blue background, but we only have one background colour

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const Pagination = ({ totalPages, padding, hitsPerPage = 5, inverted = false, ...rest }: PaginationProps) => {
+  const { refine, createURL, pages, currentRefinement, isFirstPage, isLastPage, nbPages, nbHits } = usePagination({
+    totalPages,
+    padding,
+  })
+
+  const { resultsRef } = useContext(PaginationContext)
+  const prevRefinement = useRef<number>(currentRefinement)
+  const prefersReducedMotion = usePrefersReducedMotion()
```
wenche, line 35: I think we can leave this as is for now. But just a friendly reminder about this media query and the scroll to top-animation: Animations is not actually something we have in our code base, so we should see if it fits or not.

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }, [currentRefinement])
+
+  if (!nbHits || nbHits === 0 || nbHits <= hitsPerPage) {
+    return null
+  }
+
+  return (
+    <PaginationList {...rest}>
+      <PaginationItem
+        ariaLabel="First page"
```
wenche, line 54: We should add this to text snippets

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+        refine={refine}
+        inverted={inverted}
+      >
+        <Icon data={chevron_left} />
+      </PaginationItem>
+
+      {pages.map((page) => (
+        <PaginationItem
+          key={page}
+          ariaLabel={`Page ${page + 1}`}
```
wenche, line 79: Same, add to text snippets

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+          createURL={createURL}
+          refine={refine}
+          inverted={inverted}
+        >
+          {page + 1}
+        </PaginationItem>
+      ))}
+
+      <PaginationItem
+        ariaLabel="Next"
```
wenche, line 92: Text snippet

components/ImprovedSearch/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+        isDisabled={isLastPage}
+        createURL={createURL}
+        refine={refine}
+        inverted={inverted}
+      >
+        <Icon data={chevron_right} />
+      </PaginationItem>
+
+      <PaginationItem
+        ariaLabel="Last page"
```
wenche, line 104: And same

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,93 @@
+import { usePagination } from 'react-instantsearch-hooks-web'
+import { Button } from "./Button"
+import styled from 'styled-components'
+
+
+const StyledListItem = styled.li`
```
wenche, line 6: Prefer EDS list component, as already mentioned

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  display: inline-block;
+`
+
+const PaginationLink = styled(Button)<{ isCurrent?: boolean; inverted?: boolean }>`
+  width: 44px;
```
wenche, line 11: Might consider a CSS variable here, as it seems like this should be a circle. Will make it easier to change size later on, if needed

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  height: 44px;
+  color: ${(props) => (props.inverted ? 'hsl(184, 31%, 74%)' : 'hsl(206, 32%, 21%);')}
```
wenche, line 13: We should use design tokens from EDS for the colour variables here

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ${({ isCurrent, inverted }) =>
+    isCurrent && {
```
wenche, line 15: We can remove the added complexity for inverted mode 👍

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  :disabled {
+    cursor: auto;
+    color: ${(props) => (props.inverted ? 'hsl(207, 14%, 37%)' : 'hsl(0, 0%, 86%)')};
+  }
+`
+
+export function isModifierClick(event: React.MouseEvent) {
```
wenche, line 34: We should prefer arrow function to align with the rest of the code base

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const isMiddleClick = event.button === 1
+  return Boolean(isMiddleClick || event.altKey || event.ctrlKey || event.metaKey || event.shiftKey)
+}
+
+type PaginationItemProps = React.ComponentProps<'a'> &
```
wenche, line 39: Will probably need to use EDS list item props here

wenche (changes requested):
Added some comments 👍 

One more question: In Envis only 3 numbers are visible. Like this
` firstPage prevPage number number number nextPage lastPage`
In this implementation, 5 numbers are visible. Is this a setting, or will it show number for each page? Like if we have 100 hits, will 10 numbers show? That will be an issue, if so. :) Also, even with 5 numbers, this overflows horizontally on smaller screens.

components/ImprovedSearch/PaginationItem.tsx
ShikhaMishra9: I tried using tokens.colors.interactive.primary__resting.hex but was throwing error  and i was not sure exactly which one should i use here .

wenche: Should be doable, we use ESS tokens for colours other places in the code if you search. Exactly which one to use it's always a bit hard to figure out 😅 But this one `tokens.colors.interactive.primary__resting.hex` should work, I use it for the StepperItem 🤔

components/ImprovedSearch/Pagination.tsx
```diff
@@ -33,10 +36,11 @@ export const Pagination = ({ totalPages, padding, hitsPerPage = 5, inverted = fa
   const { resultsRef } = useContext(PaginationContext)
   const prevRefinement = useRef<number>(currentRefinement)
   const prefersReducedMotion = usePrefersReducedMotion()
+  const intl = useIntl()
 
   useEffect(() => {
     if (!prefersReducedMotion && resultsRef?.current && currentRefinement !== prevRefinement.current) {
-      resultsRef.current.scrollIntoView({ behavior: 'smooth' })
```
wenche: It seems like you have some issues with the eslint setup or something? 🤔

components/ImprovedSearch/Pagination.tsx
```diff
@@ -20,11 +24,10 @@ const PaginationList = styled.ul`
 
 export type PaginationProps = {
   hitsPerPage?: number
-  inverted?: boolean
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 } & UsePaginationProps
 
-export const Pagination = ({ totalPages, padding, hitsPerPage = 5, inverted = false, ...rest }: PaginationProps) => {
```
wenche: Nice to get rid of this! 👍

components/ImprovedSearch/PaginationItem.tsx
wenche: I think I would probably done that yes. :) Like
```
--button-size: 44px;
width: var(--button-size);
height: var(--button-size);
```
But in our case here, we could probably also use the round variant of EDS? Don't they have something like that? I guess 44px is something that is needed in Envis, and not necessarily something we will need

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -1,43 +1,37 @@
...
-  display: inline-block;
-`
-
-const PaginationLink = styled(Button)<{ isCurrent?: boolean; inverted?: boolean }>`
+const PaginationLink = styled(Button)<{ isCurrent?: boolean }>`
   width: 44px;
   height: 44px;
-  color: ${(props) => (props.inverted ? 'hsl(184, 31%, 74%)' : 'hsl(206, 32%, 21%);')}
-  ${({ isCurrent, inverted }) =>
+  color: ("hsl(206, 32%, 21%);")}
```
wenche, line 8: This one was weird, you should'nt need the (" and prefer EDS token. You can also ask Espen if you don't know which colour it should be

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -1,43 +1,37 @@
...
+      color: "rgba(61, 61, 61, 1)",
     }}
-
   :hover {
-    color: ${(props) => (props.inverted ? 'vhsl(0, 0%, 0%)' : '')};
+    color: ( "")};
 
     :disabled {
-      color: ${(props) => (props.inverted ? 'hsl(207, 14%, 37%)' : 'hsl(0, 0%, 86%)')};
+      color: ("hsl(0, 0%, 86%)")};
```
wenche, line 18: Same as above

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+  width: var(--button-size);
+  height: var(--button-size); 
+  color: tokens.colors.text.static_icons__default.hex} 
+  ${({ isCurrent }) =>
+    isCurrent && {
+      background: tokens.colors.interactive.primary__resting.hex,
+      color: tokens.colors.text.static_icons__default.hex,
+    }}
+  :hover {
+    color: tokens.colors.interactive.primary__resting.hex;
```
aronhoyer, line 17: remember to interpolate tokens

components/ImprovedSearch/PaginationItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+    color: tokens.colors.ui.background__medium.hex;
+  }
+`
+const { Item: EdsItem } = List
+
+const StyledListItem = styled(EdsItem)``
+
+export const isModifierClick = (event: React.MouseEvent) => {
+  const isMiddleClick = event.button === 1
+  return Boolean(isMiddleClick || event.altKey || event.ctrlKey || event.metaKey || event.shiftKey)
```
aronhoyer, line 35: this disables default operating system behaviors. for instance, a middle mouse click on a link on windows, will open the link in a new tab. likewise with a cmd + left click on mac. i don't think it's a good idea to block operating system behaviors

locales/english.json
```diff
@@ -39,6 +39,10 @@
   "navbar.placeholderSearch": "Search",
   "navbar.ariaUserMenu": "User menu",
   "navbar.titleAccount": "Account",
+  "pagination.firstPage": "First page",
+  "pagination.previous": "Previous",
+  "pagination.next": "Next",
+  "pagination.lastPage": "Last page",
```
aronhoyer, line 45: prefix with page context. i.e. `"search"`

aronhoyer (changes requested):

wenche (approved):
Good to go!

PR #338, opened 2023-02-02
Add search statistics

aronhoyer:
Resolves [AB#97331](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/97331)


wenche (changes requested):
Looks good code wise 👍 

Should we use another variable name than pageSize btw, had to think because it was different than the rest of the algolia terminology. I'm used to hits per page. No big deal it you think this is better tho

In Figma the text is `1 – 20 of 1337 items`, should change to that I guess?

wenche (approved):
LGTM 👍

PR #340, opened 2023-02-03
Feature toggle new search

wenche:
- Added a simple feature toggle that will remove the search input field if the feature toggle is not set
- Send 404 for the new search if the feature flag is not set
- 404 needed to be hard coded because 404 does not have getServersideProps, but I don't think this is a big deal
- We have to fetch the feature flags server side because of the Next.js/Docker/Radix issue

Part of [AB#97248](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/97248)


aronhoyer (commented):
have tested locally and looks good! :+1:

would it make more sense, though, to cast flags to boolean when sent from `getServerSideProps`?

something like
```js
export const getServerSideProps = () => ({
  USE_IMPROVED_SEARCH: process.env.USE_IMPROVED_SEARCH === "true",
)}
```

wenche:
> have tested locally and looks good! 👍
> 
> would it make more sense, though, to cast flags to boolean when sent from `getServerSideProps`?
> 
> something like
> 
> ```js
> export const getServerSideProps = () => ({
>   USE_IMPROVED_SEARCH: process.env.USE_IMPROVED_SEARCH === "true",
> )}
> ```
Actually a good idea 🤔  Ble så opphengt i at det måtte være string fordi det var det som kom fra env.

Let's fix that on Monday

pages/404.tsx
```diff
@@ -46,7 +46,7 @@ const PageNotFound: NextPage = () => {
   const intl = useIntl()
 
   return (
-    <Page documentTitle={intl.formatMessage({ id: "404.documentTitle" })}>
+    <Page documentTitle={intl.formatMessage({ id: "404.documentTitle" })} useImprovedSearch>
```
aronhoyer, line 49: doesn't this set `useImprovedSearch` to `true` instead of passing the value of the prop?

wenche: Yup, should be false

aronhoyer (approved):

PR #343, opened 2023-02-06
fix: Fixes bug where url query and search box state didn't match

aronhoyer:
Fixes [AB#98082](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/98082)


wenche:
Closing this, as we want to avoid to do things ourselves since Algolia has announced a dedicated library to handle Algolia + Next.js routing

PR #345, opened 2023-02-06
Url state bug

wenche:
Experiments for stupid bug


wenche:
While waiting for Algolia's reply for the reported issue, we decided to add this fix to our code base even if it's not 100% since it is (hopefully 🤞 ) better than the original code

PR #346, opened 2023-02-07
fix: Fix bug where collibra search fails

aronhoyer:


wenche (approved):
I'm feeling lucky 😅

wenche (changes requested):
When I test this locally,  I don't get any results back even with *

Skjermbilde 2023-02-08 kl  07 47 33

aronhoyer:
@wenche should™ be fixed now

wenche (approved):
Can confirm!

PR #349, opened 2023-02-09
Changes for no filter and no results

ShikhaMishra9:


components/ImprovedSearch/Hits.tsx
```diff
@@ -28,6 +34,18 @@ type Props = {
 export const Hits = () => {
   const { hits } = useHits()
   const { query } = useSearchBox()
+
+  if (!hits || hits.length === 0) {
+    return (
+      <StyledList variant="numbered">
+        <NoRelevant>
```
wenche, line 41: It's not allowed to add a span inside an ordered list, we need another container than a list to wrap this

components/ImprovedSearch/Hits.tsx
```diff
@@ -28,6 +34,18 @@ type Props = {
...
+          <Typography group="paragraph" variant="body_short" style={{ color: "Tertiary", textAlign: "center" }}>
```
wenche, line 42: Can't add color as a string `tertiary`

components/ImprovedSearch/RefinementList.tsx
```diff
@@ -53,7 +53,9 @@ export const RefinementList = (props: RefinementListProps) => {
         </EdsProvider>
       ) : (
         <NoRelevant>
-          <FormattedMessage id="asset.no.relevant.filters" defaultMessage="No assets for this filter" />
+          <Typography group="paragraph" variant="meta" color="Tertiary">
```
wenche, line 56: If you inspect using dev tools you will see that `color="tertiary"` doesn't work

wenche (changes requested):
The "Filter" header does still not show up, did you agree with Espen to do that in another PR?

components/ImprovedSearch/Hits.tsx
ShikhaMishra9: ok 

> The "Filter" header does still not show up, did you agree with Espen to do that in another PR?

yes i thought to have it in another pat after checking with you once

components/ImprovedSearch/RefinementList.tsx
ShikhaMishra9: yeah ... i  kept it to ask ...if this is something we can use through token colors

wenche (changes requested):
Remove the StyledList :thumbsup:

wenche (approved):
Looks good now 👍

PR #352, opened 2023-02-14
 PR for clear Filter

ShikhaMishra9:


pages/search-beta.tsx
```diff
@@ -10,25 +10,33 @@ import {
...
-
-} from "components/ImprovedSearch"
+import { SearchBox, Hits, Hit, PlainRefinementList, Pagination } from "components/ImprovedSearch"
 import { Page } from "components/Page"
 import { SearchStatistics } from "components/SearchStatistics"
 import { Section } from "components/Section"
 import { searchClient, searchClientServer } from "config"
 import englishTexts from "locales/english.json"
 
+const StyledName = styled(ClearRefinements)`
```
wenche, line 25: I think this code style is pretty "off" with our code base since you need to override styles on Algolia's CSS class names. I think we should avoid this. You should rather [use the hooks approach](https://www.algolia.com/doc/api-reference/widgets/clear-refinements/react-hooks/#hook) and use an EDS button (or whatever Espen suggests). With this approach, we can control the UI with styled-components and we avoid relying on internal Algolia class names 👍  This is also the way we do it with the rest of the components

pages/search-beta.tsx
```diff
@@ -122,8 +134,29 @@ const Search = ({ serverState, isServerRendered, serverUrl }: Props) => (
             {/* @ts-ignore  */}
             <Hits hitComponent={Hit} />
           </StyledHits>
+
           <FilterContainer>
```
wenche, line 138: I  would have refactored the FilterContainer to its own file to avoid to much components with styles in this file

pages/search-beta.tsx
```diff
@@ -122,8 +134,29 @@ const Search = ({ serverState, isServerRendered, serverUrl }: Props) => (
...
+                  button: "MyCustomClearRefinementsButton",
+                  disabledButton: "DisabledButton",
+                }}
+              >
+                <FormattedMessage id="improvedSearch.filter.clear" />
+              </StyledName>
+            </Header>
+
+            <hr />
+            <PlainRefinementList label="Data Office" attribute="community" />
```
wenche, line 158: Should rename this component :)

wenche (changes requested):
Hi Shikha, some good improvements here. I would suggest to use the hooks approach with EDS component instead of the ready made Algolia component to better align this with our code base

pages/search-beta.tsx
ShikhaMishra9: This one when I tried moving to Filter component the whole CSS had turbulence . Can open it as separate task and then move all containers to respective components

components/ImprovedSearch/CustomClearRefinement.tsx
```diff
@@ -0,0 +1,41 @@
+import { Button } from "@equinor/eds-core-react"
+import { tokens } from "@equinor/eds-tokens"
+import React from "react"
+import { useClearRefinements, UseClearRefinementsProps } from "react-instantsearch-hooks-web"
+import { useIntl } from "react-intl"
+import styled from "styled-components"
+
+const ButtonNew = styled(Button)`
```
wenche, line 8: Perhaps change the name?

components/ImprovedSearch/CustomClearRefinement.tsx
```diff
@@ -0,0 +1,41 @@
...
+  color: ${tokens.colors.text.static_icons__default.hex};
+  background: ${tokens.colors.infographic.primary__moss_green_13.hex};
+  border-width: 0;
+  font-weight: 14pt;
+  padding-left: 8.15rem;
+  :hover {
```
wenche, line 14: Did Espen tell you to override the hover effect from EDS? 🤔  I think it is strange that it does not have the hover effect. Should it not be a ghost button like the ones in the header instead?

components/ImprovedSearch/CustomClearRefinement.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { tokens } from "@equinor/eds-tokens"
+import React from "react"
+import { useClearRefinements, UseClearRefinementsProps } from "react-instantsearch-hooks-web"
+import { useIntl } from "react-intl"
+import styled from "styled-components"
+
+const ButtonNew = styled(Button)`
+  color: ${tokens.colors.text.static_icons__default.hex};
+  background: ${tokens.colors.infographic.primary__moss_green_13.hex};
+  border-width: 0;
```
wenche, line 11: If you use the ghost variant, you should not need to override the border

components/ImprovedSearch/CustomClearRefinement.tsx
```diff
@@ -0,0 +1,41 @@
...
+  font-weight: 14pt;
```
wenche, line 12: You should prefer rem (or px)

components/ImprovedSearch/CustomClearRefinement.tsx
```diff
@@ -0,0 +1,41 @@
...
+  padding-left: 8.15rem;
```
wenche, line 13: You can't do this to position the button. 😄  If you tab through it or inspect it, you will notice that it doesn't look good. I would suggest to rather use `justify-content: space-between` on the header flex box instead
Skjermbilde 2023-02-16 kl  10 00 38
Skjermbilde 2023-02-16 kl  10 00 55


Skjermbilde 2023-02-16 kl  10 01 40

components/ImprovedSearch/CustomClearRefinement.tsx
```diff
@@ -0,0 +1,41 @@
...
+    color: ${tokens.colors.text.static_icons__default.hex};
+    background: ${tokens.colors.infographic.primary__moss_green_13.hex};
+    border-width: 0;
+  }
+`
+
+const ButtonDisabled = styled(Button)`
+  color: ${tokens.colors.text.static_icons__tertiary.hex};
+  background: ${tokens.colors.infographic.primary__moss_green_13.hex};
+  border-width: 0;
```
wenche, line 24: Same comments as above.

And also a lot of duplicate code here. A better approach would be one of the following:
- Use styled components prop to send in isDisabled to the disabled button and [use conditional styles](https://styled-components.com/docs/basics#adapting-based-on-props)
- Use conditional styles with CSS variable to the component like[ I did here](https://github.com/equinor/data-marketplace/blob/main/htmlParsing/portableText/components/Image.tsx). This might be closer to the Envis approach tho, but just for the record
- [Extend styles with styled components](https://styled-components.com/docs/basics#extending-styles)

wenche (changes requested):
Some comments added

And I also found a bug or a strange issue:

-  Go to page 3 with the pagination and click Clear Refinement -> the pagination info in the URL is gone and your are back to page 1

I would not have expected this to happen? 🤔  What do you think @aronhoyer

aronhoyer:
> What do you think @aronhoyer

i guess that's unexpected to an extent, but at the same time it kind of makes sense. the list of hits are different with and without refinements (naturally). so the results on page 3 would potentially be drastically different once you clear the refinements.

components/ImprovedSearch/CustomClearRefinement.tsx
ShikhaMishra9: actually the color coming by default from hiver is not what we  wanted so i have to use it in styling and get the color of our choice

wenche: OK, it still feels a little bit weird that we don't have a hover effect on this button but if that's what Espen wants 🤷‍♀️

wenche (approved):
LGTM 👍 

Nice with just one button :)

PR #356, opened 2023-02-20
changed font for search results

ShikhaMishra9:


components/ImprovedSearch/Hit.tsx
```diff
@@ -94,12 +94,16 @@ export const Hit = ({ hit, hasQuery }: HitProps) => {
     <StyledLink href={{ pathname: "/assets/[id]", query: { id } }}>
       {community &&
         community.map((item) => (
-          <Typography variant="overline" key={item}>
+          <Typography variant="overline" style={{ fontSize: "0.7rem", fontWeight: "medium" }} key={item}>
```
wenche, line 97: This is a really strange font size? Is it correct? `11.2px`

This should also be rendered as a `span` instead of `p`.  I probably forgot that when it was added

components/ImprovedSearch/Hit.tsx
```diff
@@ -109,7 +113,7 @@ export const Hit = ({ hit, hasQuery }: HitProps) => {
         />
       </Typography>
       {hasQuery ? (
-        <StyledTypography variant="body_short">
+        <StyledTypography variant="body_short" style={{ fontSize: "0.87rem", fontWeight: "regular" }}>
```
wenche, line 116: Same here, this is a really weird font size? `13.92px`. And did you agree upon that the font size should change only if there's a query? 🤔

wenche (changes requested):

components/ImprovedSearch/Hit.tsx
ShikhaMishra9: In both actually ..I am updating

wenche: Looks good! Had a live demo with Espen and we agreed upon that 14px was a little bit too small so I'm changing it back to 16

wenche (approved):
LGTM to Espen and me 👍

PR #361, opened 2023-02-28
update responsibility table

ShikhaMishra9:


wenche (changes requested):
I think we should not just remove the tables, but we should also use the filter option in the `/api/assets/{id}/maintainers` to just fetch the maintainer groups we need

wenche (approved):
LGTM 👍

PR #363, opened 2023-03-02
✨ Add browse specific business area

wenche:
Resolves [AB#102013](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/102013)


wenche:
Felt like it was overkill to add the names in the lang.json file, but we could of course do it and reference to the string in the data array

PR #365, opened 2023-03-06
Add coloured backgrounds for asset detail and checkout

wenche:
Resolves [AB#102023](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/102023)


ShikhaMishra9 (approved):
LGTM ....Using container does makes it better

PR #368, opened 2023-03-07
✨ Use the real Algolia index name

wenche:
Resolves ab#102991

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -68,12 +68,13 @@ const InfoBanner = styled.div`
 `
 
 type Props = {
+  algoliaIndexName: string
   featureFlags: {
     USE_IMPROVED_SEARCH: boolean
   }
 }
 
-const Frontpage: NextPage<Props> = ({ featureFlags = { USE_IMPROVED_SEARCH: false } }) => {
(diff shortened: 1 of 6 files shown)


pages/index.tsx
```diff
@@ -112,8 +113,15 @@ export default Frontpage
 export const getServerSideProps: GetServerSideProps = async () => {
   const USE_IMPROVED_SEARCH = process.env.USE_IMPROVED_SEARCH === "true"
 
+  const indexName = process.env.ALGOLIA_SEARCH_INDEX ?? ""
+
+  if (indexName === "") {
+    console.log("Missing the Algolia search index name")
```
aronhoyer, line 119: we should probably handle this case in the UI somehow. can probably do that in a later pr

aronhoyer (commented):
tested locally and looks awesome! y'all have done an amazing job 👏 

real quick though, should `/search-beta` be retired and moved to `/search`?

aronhoyer (approved):
previous comment resolved in slack conversation

wenche:
Yes, we have quite a bit of a clean up job to do but let's take it to production first, and then remove and rename all the other stuffs

PR #370, opened 2023-03-09
Remove the old search page

wenche:
Resolves ab#103203


nilsml (approved):
As far as I can see it looks ok, but a hard one to review without actually testing. Need to get the ENV set up on my Codespace for next time.

PR #371, opened 2023-03-09
Upgrade packages

wenche:
Resolves [AB#103212](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/103212)

Will need to add the new package for Next + Algolia before we can update the Next.js stuff


nilsml (approved):
LGTM

PR #372, opened 2023-03-10
⬆️ Upgrade to released version of the Algolia and Next lib

wenche:
Resolves ab#103433


wenche:
I have one thing here that I would like to discuss on Monday, but it doesn't prevent us from merging this PR now 👍

ShikhaMishra9 (approved):
looks good

PR #374, opened 2023-03-13
⚗ Test for using link as pagination item

wenche:
Experiments all the way


ShikhaMishra9 (approved):
For now this is ok... we will change background of current page on mouse hover once EDS issue is fixed

PR #375, opened 2023-03-15
Spike for Next 13's app folder

wenche:
Part of [AB#102022](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/102022)

This is a POC of concept for using the app folder. I'm facing some issues:
1.  I tried the new font optimisation feature with local font, so I downloaded the EDS variable font from the CDN. The best approach  for using the font is, from my point of view, [the featured example from the docs](https://beta.nextjs.org/docs/optimizing/fonts#local-fonts) where the font is appended to the `html` element like this
`<html class="__className_b4df81">` 
```
.__className_b4df81 {
    font-family: '__equinorFont_b4df81', '__equinorFont_Fallback_b4df81';
}
```
But that gives some challenges with the EDS components since every EDS component has  the CSS rule `font-family: Equinor` which will override the one added by Next.js :/

2. `<GlobalStyles />` in the layout folder gives `error - Error: React.Children.only expected to receive a single React element child.` server side. But everything seems to work anyway, and I have not found any ways of adding global styles 🤷‍♀️ 

3.I tried the dynamic SEO approach from 13.2, doesn't work 100% with `react-intl` because we can't use the `useIntl()` hook in this function. But for our use case with just a single language I think it's OK with the simple workaround I did

4. With the feature toggle I'm using for this we will see the default Next.js error page if you go to /test with the feature off. I don't think this is critical

To actually port the rest of the pages will require a lot of work and we should probably wait  a little bit here. But I think it should be doable, but I don't know how much for instance the Router part will take (the index page doesn't actually use the router)


ShikhaMishra9 (approved):
Looks good .....looking forwward for new features we can use with Next13.

PR #378, opened 2023-03-17
Use CSS vars for colours

wenche:
Part of [AB#103060](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/103060)

In order to go through our colours to create a colour palette, we decided to put all the colours as CSS variables for an easy overview


wenche:
Since EDS is in the middle of a big refactoring for colours, I just added the colours as hsla. This will maximise DX in our application. When EDS 2.0 is out, we should revisit the colours and see if we should use EDS again

The complete list of colours:
Skjermbilde 2023-03-17 kl  14 05 57

ShikhaMishra9 (approved):
Looks good to get control over colors :)

PR #382, opened 2023-03-27
Added New tag for data products which are created within one month

ShikhaMishra9:
Part of [AB#105363](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/105363)

### DIFF
--- a/components/Search/Hit.tsx
+++ b/components/Search/Hit.tsx
@@ -58,7 +58,7 @@ const TopContainer = styled.div`
 `
 
 const StyledChip = styled(Chip)`
-  background-color: var(--tag-outline);
+  background-color: var(--disappointing-vanilla);
   color: var(--charcoal);
 `
 
@@ -66,7 +66,7 @@ const StyledChipNew = styled(Chip)`
   background-color: var(--white);
   color: var(--charcoal);
   text-transform: uppercase;
-  border: 2px solid var(--tag-outline);
+  border: 2px solid var(--disappointing-vanilla);
 `
 
 const StyledTypography = styled(Typography)`
--- a/components/Search/Hit.tsx
+++ b/components/Search/Hit.tsx
@@ -58,14 +58,15 @@ const TopContainer = styled.div`
 `
 
 const StyledChip = styled(Chip)`
-  background-color: rgb(255, 218, 168);
+  background-color: var(--tag-outline);
   color: var(--charcoal);
 `
 
 const StyledChipNew = styled(Chip)`
   background-color: var(--white);
   color: var(--charcoal);
-  border: 2px solid rgb(255, 218, 168);
+  text-transform: uppercase;
+  border: 2px solid var(--tag-outline);
 `
 
 const StyledTypography = styled(Typography)`
--- a/components/Search/Hit.tsx
+++ b/components/Search/Hit.tsx
@@ -52,7 +52,6 @@ const TagsContainer = styled.div`
 `
 
 const TopContainer = styled.div`
-  margin-top: ${tokens.spacings.comfortable.medium};
   display: flex;
   justify-content: space-between;
   gap: ${tokens.spacings.comfortable.large};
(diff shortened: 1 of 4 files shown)


components/Search/Hit.tsx
```diff
@@ -50,11 +51,24 @@ const TagsContainer = styled.div`
   gap: ${tokens.spacings.comfortable.small};
 `
 
+const TopContainer = styled.div`
+  margin-top: ${tokens.spacings.comfortable.medium};
```
aronhoyer, line 55: where did you get this margin from? from the screen shot your shared in slack, it seems like only cards with a "new" tag get a lot of top margin. i suspect this line is the culprit, and i think it can be removed

components/Search/Hit.tsx
```diff
@@ -77,22 +93,80 @@ If neither the excerpt or the description is a match (like the match can be for
...
+        {tags.length > 0 && (
+          <TagsContainer>
+            {tags.map((item) => (
+              <StyledChip key={item}>{item}</StyledChip>
+            ))}
+          </TagsContainer>
+        )}
+      </StyledLink>
+    )
+  }
```
aronhoyer, line 169: i would strongly suggest, instead of rendering the entire component with a new tag if it's a new data product, that you render the new tag if it's a new data product. the difference being:
```js
// instead of
if (isNew()) {
  // entire component
}
// we do
return (
  <StyledLink>
    <FlexContainer>
      {community && <Community />}
      {isNew() && <NewTag />}
    </FlexContainer>
  </StyledLink>
)
```
doing this means that if we need to make a change to the hit card, we don't need to do it two places :+1:

components/Search/Hit.tsx
```diff
@@ -77,22 +93,80 @@ If neither the excerpt or the description is a match (like the match can be for
...
+
   return hit.excerpt ? "excerpt" : "description"
 }
 
+const isNewDataProduct = (hit: HitProps["hit"]) => {
+  const currentDate = new Date()
+  const compareDate = new Date(currentDate.setMonth(currentDate.getMonth() - 1))
+  const createdAtDate = new Date(hit.createdAt)
+  return createdAtDate.getTime() >= compareDate.getTime()
+}
```
aronhoyer, line 118: i don't think this should be calculated in the client. at least in its current state. this implementation will use the client's time zone, while the `createdAt` date is likely UTC, so people in different time zones will see different results. which for this use case isn't the biggest deal, but we should strive for every user to get presented the same results regardless of time zone imho

aronhoyer: additionally, if we implemented it in the indexer service, we would be able to let the users filter for it :+1:

components/Search/Hit.tsx
```diff
@@ -50,11 +51,23 @@ const TagsContainer = styled.div`
...
+
 const StyledChip = styled(Chip)`
   background-color: rgb(255, 218, 168);
   color: var(--charcoal);
 `
 
+const StyledChipNew = styled(Chip)`
+  background-color: var(--white);
+  color: var(--charcoal);
+  border: 2px solid rgb(255, 218, 168);
```
wenche, line 68: Please don't add colours like this since all other colours are now as CSS variables

locales/english.json
```diff
@@ -38,6 +38,7 @@
   "search.box.placeholder": "Search for data...",
   "search.header": "Search for the data you need",
   "search.title": "Search for assets",
+  "search.results.new.dataproduct": "NEW",
```
wenche, line 41: I would prefer to write it normally, and rather use CSS for uppercase

wenche (changes requested):

components/Search/Hit.tsx
wenche: Do we have another date then createdTime? Since I don't think this is as accurate as we need? 🤔  Like country data, it has a create data several months ago even if it went live the other day

components/Search/Hit.tsx
ShikhaMishra9: This color did not exist in CSS variable. Have added it now and changed in Styled chip as well

ShikhaMishra9: Done

styles/globals.ts
```diff
@@ -80,6 +80,7 @@ a {
   --warning: hsla(25,100%,92%,1); /* tokens.colors.interactive.warning__highlight.hsla */
   --transparent-black: hsla(0, 0%, 0%, 0.14); /* Not an EDS colour */
   --warning-icon-fill: hsla(34,100%,33.9%,1); /* tokens.colors.interactive.warning__text.hsla */
+  --tag-outline: hsla(34.483, 100%, 83%, 1);
```
wenche, line 83: Should remove this

components/Search/RefinementList.tsx
```diff
@@ -43,6 +43,7 @@ const StyledLabel = styled.label`
 const LabelName = styled.span`
   white-space: nowrap;
   overflow-x: hidden;
+  line-height: 1.2rem;
```
wenche, line 46: We need to add the line height on both the text and number, the number is a little bit off now
Skjermbilde 2023-03-27 kl  12 45 27

components/Search/Hit.tsx
```diff
@@ -50,9 +51,22 @@ const TagsContainer = styled.div`
...
-  background-color: rgb(255, 218, 168);
+  background-color: var(--disappointing-vanilla);
+  color: var(--charcoal);
+`
+
+const StyledChipNew = styled(Chip)`
+  background-color: var(--white);
   color: var(--charcoal);
+  text-transform: uppercase;
+  border: 2px solid var(--disappointing-vanilla);
```
wenche, line 69: In Figma the tag has a height of 24px. Here's an example of hard coding the height to 24px (should have used rems instead) and a line height of 1 (the same as the text size since it's all uppercase). I think the height of 22px for EDS Chip is because EDS uses `content-box` and not `border-box` for box-sizing. Would usually not suggested a hard-coded height, but in this particular case it does look better 🤔
Skjermbilde 2023-03-27 kl  13 06 14

components/Search/Hit.tsx
```diff
@@ -77,32 +93,49 @@ If neither the excerpt or the description is a match (like the match can be for
...
-
+      <TopContainer>
+        {community &&
+          community.map((item) => (
+            <Typography variant="overline" style={{ fontSize: "0.75rem", fontWeight: "medium" }} key={item}>
+              {item}
+            </Typography>
+          ))}
+        {isNewDataProduct(hit) && (
+          <StyledChipNew style={{ fontSize: "0.75rem", fontWeight: "medium" }}>
```
wenche, line 134: Medium is not a known value for font weight, but EDS adds 500 and that's the same as Figma. Should remove this

wenche (changes requested):

wenche (approved):
LGMT 👍 

I think it looks good now.

A final note about line height: My example used 1,  not 1rem. The unit less way of setting the line height is in most cases the preferred one :) Since it will use the font size and not inheritance. In this particular case it's OK with 1rem since it is less then 20 (24px- 2*2px border)

PR #397, opened 2023-04-21
refactor: Complete front page next 13 port

aronhoyer:
I tried to find a way to nest the front page and its components within a directory (like `app/page/index.ts`), but didn't find a way to do it that worked.

The current solution works, but it kind of crowds the `app` directory.

Resolves [AB#108464](https://dev.azure.com/EquinorASA/d02c350d-092d-41bf-a812-f05d5d8ad1b0/_workitems/edit/108464)


wenche (changes requested):
Looks good, but I see what you mean about the structure here 🤔  Might consider to put the components for the front page somewhere else, let's revisit when everything is migrated

You need to update the CSS variables since this PR now reintroduces some of the deleted ones 👍

